PR #3, opened 2023-04-19
Update readme for node 16

gregegan:


CassOnMars (commented):
This project is not a NodeJS project – node is not required

PR #38, opened 2024-01-19
feat: add balance flag to CLI 

### LINKED ISSUE #37: wallet balance from node command.
Request Type : Feature/Enahancement.

Currently the `node` binary has the below options, however having `-get-balance`   feature will help to get the balance details vs 
 `-db-console` and looking at the right bottom  screen for balances. 
 This option should able to give straight forward response for operators query to get current wallet balance.

```
~/ceremonyclient/node$ ./node --help
Usage of ./node:
  -config string
        the configuration directory (default ".config")
  -db-console
        starts the node in database console mode
  -import-priv-key string
        creates a new config using a specific key from the phase one ceremony
  -peer-id
        print the peer id to stdout from the config and exit
```

agostbiro:
Adds a `-balance` flag to the CLI to print the confirmed token balance. Fixes #37 .

I refactored a bit to be able to reuse the DB console's implementation for fetching the token balance.

#### Example

Example usage from the `node` directory: 

`GOEXPERIMENT=arenas go run ./... -balance`

Example output:

`Confirmed balance: 1234 QUIL`

#### Testing

I tested the command on a running node and also verified that the DB console starts and displays the correct balance after the change. 

I didn't test the path that depends on the value of the `grpcWarn` variable as I'm not sure how.

### DIFF
--- a/node/app/db_console.go
+++ b/node/app/db_console.go
@@ -841,7 +841,7 @@ func (c *DBConsole) Run() {
 	}
 	defer conn.Close()
 
-	grpcWarn := conn.Target() == defaultGrpcAddress
+	grpcWarn := c.nodeConfig.ListenGRPCMultiaddr == ""
 
 	p := tea.NewProgram(consoleModel(conn, c.nodeConfig, grpcWarn))
 	if _, err := p.Run(); err != nil {
--- a/node/app/db_console.go
+++ b/node/app/db_console.go
@@ -110,16 +110,11 @@ func (m model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	if m.conn.GetState() == connectivity.Ready {
 		if m.lastChecked < (time.Now().UnixMilli() - 10_000) {
 			m.lastChecked = time.Now().UnixMilli()
-			info, err := m.client.GetTokenInfo(
-				context.Background(),
-				&protobufs.GetTokenInfoRequest{},
-			)
+
+			tokenBalance, err := FetchTokenBalance(m.client)
 			if err == nil {
-				conversionFactor, _ := new(big.Int).SetString("1DCD65000", 16)
-				m.owned = new(big.Int).SetBytes(info.OwnedTokens)
-				m.owned.Div(m.owned, conversionFactor)
-				m.unconfirmedOwned = new(big.Int).SetBytes(info.UnconfirmedOwnedTokens)
-				m.unconfirmedOwned.Div(m.unconfirmedOwned, conversionFactor)
+				m.owned = tokenBalance.Owned
+				m.unconfirmedOwned = tokenBalance.UnconfirmedOwned
 			}
 		}
 	}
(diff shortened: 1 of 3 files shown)


node/app/db_console.go
```diff
@@ -808,11 +803,46 @@ func (c *DBConsole) Run() {
...
+
+// Runs the DB console
+func (c *DBConsole) Run() {
+	conn, err := ConnectToNode(c.nodeConfig)
+	if err != nil {
+		panic(err)
+	}
 	defer conn.Close()
+
+	grpcWarn := conn.Target() == defaultGrpcAddress
```
agostbiro, line 844: This check is not equivalent to the prior version if the user's grpc address in the config is `localhost:8337`. Let me know please if I should change it to match the prior logic exactly

agostbiro: Alternative could be to do this:

```
grpcWarn := c.nodeConfig.ListenGRPCMultiaddr != ""
```

This wouldn't match the prior logic either exactly since it doesn't check if the config value is correct, but maybe it's closer to the original intention which I don't fully understand tbh 😅

agostbiro: Ok I think I got it now: if `grpcWarn` is true, we want to display the warning `"gRPC Not Enabled, Please Configure"` and otherwise show the connection status message.

https://github.com/QuilibriumNetwork/ceremonyclient/blob/7a113ebb785907eed266a6f126f4f6eecc9cb82f/node/app/db_console.go#L464-L478

So the purpose is to warn the user if they haven't set the necessary config field yet. We can detect this if `nodeConfig.ListenGRPCMultiaddr == ""`.

I fixed the db console in [7a113eb](https://github.com/QuilibriumNetwork/ceremonyclient/pull/38/commits/7a113ebb785907eed266a6f126f4f6eecc9cb82f) and added the same warning to the CLI in [bac0344](https://github.com/QuilibriumNetwork/ceremonyclient/pull/38/commits/bac034450ce585e7528a44820053ea0eb8e3d23f).

I retested the `-balance` flag and the balance in the DB console on a node with gRPC enabled. I also tested that the warnings are shown for a node with gRPC disabled in both the CLI and the DB console.

tekinsercan:
@CassOnMars @agostbiro

I'm reaching out to request your assistance with an issue I've encountered while querying the balance on our node, which is running in a Docker container.

Despite confirming successful network connectivity with the following output:
```
/opt/ceremonyclient/node # nc -zv 0.0.0.0 8337
0.0.0.0 (0.0.0.0:8337) open
/opt/ceremonyclient/node #
```
Additionally, using the `--db-console` option indicates a `CONNECTED` status:
```
STATUS  CONNECTED
```
I'm facing an issue when attempting to execute the balance command:
```
/opt/ceremonyclient/node # go run ./... -balance
panic: error getting token info: rpc error: code = Unknown desc = get token info: get highest candidate data clock frame: item not found

goroutine 1 [running]:
main.main()
	/opt/ceremonyclient/node/main.go:101 +0x745
exit status 2
/opt/ceremonyclient/node #
```
Could you please advise if there's a specific configuration or step I might be missing?
Thank you!

PR #46, opened 2024-02-16
add basic Docker configuration

mscurtescu:
Adding both a `Dockerfile` and `docker-compose.yml`  files that allow you to easily create an image and run it.

```shell
docker build -t quilibrium -t quilibrium:1.2.5 .
docker compose up -d
```

Notes:

* a golang 1.20 base image, based on Alpine linux is used
* a docker volume is used for the `node/.config` folder (and this includes the nested `storage` subfolder)
* the p2p port 8336/udp is mapped to the host, all interfaces
* the gRPC and REST ports 8337/tcp and 8338/tcp are mapped to host, localhost interface only

In order to drop into a shell inside the container:
```shell
docker compose exec -it node sh
```

Future work:

* set defaults for `listenRESTMultiaddr` and `listenGrpcMultiaddr`
* add `HEALTHCHECK` to the image
* add monitoring related services (Prometheus, Grafana, etc) to the compose

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,11 @@
 FROM golang:1.20-alpine
 
+LABEL org.opencontainers.image.title="Quilibrium Network Node"
+LABEL org.opencontainers.image.description="Quilibrium is a decentralized alternative to platform as a service providers."
+LABEL org.opencontainers.image.vendor=Quilibrium
+LABEL org.opencontainers.image.url=https://quilibrium.com/
+LABEL org.opencontainers.image.documentation=https://quilibrium.com/docs
+
 ENV GOEXPERIMENT=arenas
 
 WORKDIR /opt/ceremonyclient
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.20-alpine
+FROM golang:1.20.14-alpine3.19
 
 LABEL org.opencontainers.image.title="Quilibrium Network Node"
 LABEL org.opencontainers.image.description="Quilibrium is a decentralized alternative to platform as a service providers."
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM golang:1.20
+
+ENV GOEXPERIMENT=arenas
+
+WORKDIR /opt/ceremonyclient
+
+COPY . . 
+
+WORKDIR /opt/ceremonyclient/node
+
+RUN go mod download && go mod verify
+
+CMD ["go", "run", "./..."]
+
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.20
+FROM golang:1.20-alpine
 
 ENV GOEXPERIMENT=arenas
 
@@ -9,6 +9,6 @@ COPY . .
 WORKDIR /opt/ceremonyclient/node
 
 RUN go mod download && go mod verify
+RUN go build ./...
 
-CMD ["go", "run", "./..."]
-
+ENTRYPOINT ["go", "run", "./..."]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,14 @@
+version: "3.8"
+
+name: quilibrium
+
+volumes:
+  config:
+
+services:
+  node:
+    build: ./
+    image: quilibrium
+    restart: always
+    volumes:
+      - config:/opt/ceremonyclient/node/.config
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,9 +2,6 @@ version: "3.8"
 
 name: quilibrium
 
-volumes:
-  config:
-
 services:
   node:
     build: ./
@@ -15,7 +12,7 @@ services:
       - '127.0.0.1:8337:8337/tcp' # gRPC
       - '127.0.0.1:8338:8338/tcp' # REST
     volumes:
-      - config:/opt/ceremonyclient/node/.config
+      - ./.config:/opt/ceremonyclient/node/.config
     logging:
       driver: "json-file"
       options:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,5 +10,9 @@ services:
     build: ./
     image: quilibrium
     restart: always
+    ports:
+      - '8336:8336/udp' # p2p
+      - '127.0.0.1:8337:8337/tcp' # gRPC
+      - '127.0.0.1:8338:8338/tcp' # REST
     volumes:
       - config:/opt/ceremonyclient/node/.config
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,3 +16,8 @@ services:
       - '127.0.0.1:8338:8338/tcp' # REST
     volumes:
       - config:/opt/ceremonyclient/node/.config
+    logging:
+      driver: "json-file"
+      options:
+        max-file: "5"
+        max-size: 2048m


AvAcalho:
Thank you for sharing this! I want to open a healthy discussion on the the setup because I had this doubts myself... Volume config is persistent but if volume is gone `keys.yml `and `config.yml` are gone forever.
Should these 2 files be mapped to the host to be easier to manage them? If so, @CassOnMars can we have a `keys.yml` and `config,yml `generator routine that besides generate them if they don't exist, also generate them if they exist but are empty ? 
I ask this because docker will generate a folder (not a file) if we don't have a file on the mapped volume path

Perksey:
Haha! I was just coming to make a little dockerized node, glad someone beat me to it!

mscurtescu:
> Thank you for sharing this! I want to open a healthy discussion on the the setup because I had this doubts myself... Volume config is persistent but if volume is gone `keys.yml `and `config.yml` are gone forever. Should these 2 files be mapped to the host to be easier to manage them? If so, @CassOnMars can we have a `keys.yml` and `config,yml `generator routine that besides generate them if they don't exist, also generate them if they exist but are empty ? I ask this because docker will generate a folder (not a file) if we don't have a file on the mapped volume path

I guess either way you want to backup the configuration.

The `.config` folder could be mapped to a local folder. Currently the challenge with that is that the `store` subfolder is under `.config` and ideally you really want this to be managed by docker. Docker could be configured to store volumes on a different larger and faster drive. So I think best would be to have config and storage use different folders (and not implicitly nested).

+1 for a config file generator, so they can be configured ahead of time.

Is `keys.yml` holding the identity of the node? When and how is that generated? I saw reference to an import flag.

CassOnMars (dismissed):

CassOnMars:
looks great, we do need commits signed though before it can be merged 🙏🏻

0xluk:
As far as I know `listenRESTMultiaddr` and `listenGrpcMultiaddr` are blank in the default config so we should also modify their default values or modifying the config on the fly in the entrypoint, otherwise rpc_server will not be created/exposed.

AvAcalho:
> As far as I know `listenRESTMultiaddr` and `listenGrpcMultiaddr` are blank in the default config so we should also modify their default values or modifying the config on the fly in the entrypoint, otherwise rpc_server will not be created/exposed.

We need to manually add them to `config.yml`. However this docker file is exposing the ports already by default, which makes sense to me. The only thing that user needs to do is add the entries to `config.yml`

@mscurtescu as mentioned on TG channel I'm leaving the config I'm using so you can have a look

```
# syntax=docker/dockerfile:1

FROM golang:1.20.11

# Set destination for COPY
WORKDIR /quilibrium

COPY entrypoint.sh entrypoint.sh
RUN chmod +x entrypoint.sh

# Build
RUN git clone https://github.com/QuilibriumNetwork/ceremonyclient.git

# Optional:
# To bind to a TCP port, runtime parameters must be supplied to the docker command.
# But we can document in the Dockerfile what ports
# the application is going to listen on by default.
# https://docs.docker.com/engine/reference/builder/#expose
EXPOSE 8336/udp

# Run
ENTRYPOINT ["/quilibrium/entrypoint.sh"]
CMD ["sh"]
```


```
version: '3'
services:
  quilibrium:
    image: avacalho/quilibrium:1.2.6
    container_name: quilibrium
    restart: unless-stopped
    ports:
      - '8336:8336/udp'
    volumes:
      - go_cache:/go
      - './.config/keys.yml:/quilibrium/ceremonyclient/node/.config/keys.yml:ro'
      - './.config/config.yml:/quilibrium/ceremonyclient/node/.config/config.yml:ro'
    logging:
      driver: "json-file"
      options:
        max-file: "5"
        max-size: 2048m
volumes:
  go_cache:

```

My rational above is:

- Run it with the mapped files commented for the first time (in order to generate the files in case you don't have them already)
- Extract the files or it's content with docker exec/attach or whatever is better for you
- Create the files on ./.config/ (in the host, side by side with docker-compose.yml)
- Stop the container, uncomment the lines for file mapping, start the container

mscurtescu:
> As far as I know `listenRESTMultiaddr` and `listenGrpcMultiaddr` are blank in the default config so we should also modify their default values or modifying the config on the fly in the entrypoint, otherwise rpc_server will not be created/exposed.

Yes, I was looking into that. I will follow up with another PR to sort this out. Added a note to the description in this regard.

mscurtescu:
> > As far as I know `listenRESTMultiaddr` and `listenGrpcMultiaddr` are blank in the default config so we should also modify their default values or modifying the config on the fly in the entrypoint, otherwise rpc_server will not be created/exposed.
> 
> We need to manually add them to `config.yml`. However this docker file is exposing the ports already by default, which makes sense to me. The only thing that user needs to do is add the entries to `config.yml`

As I mentioned in the previous comment, I will get back to this. Added a note in the description.


> 
> @mscurtescu as mentioned on TG channel I'm leaving the config I'm using so you can have a look
> 
> ### dockerfile
> ```
> # syntax=docker/dockerfile:1
> 
> FROM golang:1.20.11

Good point about fully specifying the version (1.20.14 instead of 1.20).

> 
> # Set destination for COPY
> WORKDIR /quilibrium
> 
> COPY entrypoint.sh entrypoint.sh
> RUN chmod +x entrypoint.sh
> 
> # Build
> RUN git clone https://github.com/QuilibriumNetwork/ceremonyclient.git

I went the other way around, assuming the git folder on the dev machine where the image is generated and copying it into the image. I guess both have pros and cons, my thinking was:

* the git folder in the dev machine can be checked out to a specific tag, so an exact version is released
* the image can use a slimmer base OS like Alpine and might not even have git installed


> 
> # Optional:
> # To bind to a TCP port, runtime parameters must be supplied to the docker command.
> # But we can document in the Dockerfile what ports
> # the application is going to listen on by default.
> # https://docs.docker.com/engine/reference/builder/#expose
> EXPOSE 8336/udp
> 
> # Run
> ENTRYPOINT ["/quilibrium/entrypoint.sh"]
> CMD ["sh"]

Is using a script instead of the direct go command better?


> ```
> 
> ### docker-compose.yml
> ```
> version: '3'
> services:
>   quilibrium:
>     image: avacalho/quilibrium:1.2.6
>     container_name: quilibrium
>     restart: unless-stopped
>     ports:
>       - '8336:8336/udp'
>     volumes:
>       - go_cache:/go

Interesting idea with the go cache. Currently when creating the image it is also building everything, maybe this is not optimal. I will revisit. With all go code prebuilt images are quite a bit larger, but I assume that containers start faster. Maybe using a cache as you are doing is better.


>       - './.config/keys.yml:/quilibrium/ceremonyclient/node/.config/keys.yml:ro'
>       - './.config/config.yml:/quilibrium/ceremonyclient/node/.config/config.yml:ro'

I think you are right about mapping the config files to a local folder. For now I will map the whole `.config` folder. We should revisit the `store` subfolder, that could easily live on a volume.

>     logging:
>       driver: "json-file"
>       options:
>         max-file: "5"
>         max-size: 2048m

Good idea with specifying logging parameters, adding this.

> volumes:
>   go_cache:
> ```
> 
> ### entrypoint.sh
> ```
> #!/bin/bash
> 
> echo "Starting Quilibrium node..."
> cd /quilibrium/ceremonyclient/node
> GOEXPERIMENT=arenas go run ./...

Note: added in the image as a global env.

> ```
> 
> My rational above is:
> 
> * Run it with the mapped files commented for the first time (in order to generate the files in case you don't have them already)
> * Extract the files or it's content with docker exec/attach or whatever is better for you
> * Create the files on ./.config/ (in the host, side by side with docker-compose.yml)
> * Stop the container, uncomment the lines for file mapping, start the container

Got it, thanks!

mscurtescu:
> looks great, we do need commits signed though before it can be merged 🙏🏻

@CassOnMars can you please review again? Made a few small changes based on feedback.

0xm1kr:
For anyone else who might be looking for the exact steps to get this running with REST / grpc:

Build and start

```sh
docker build -t quilibrium -t quilibrium:1.2.5 .
docker compose up -d
```

Connect to the container

```sh
docker compose exec -it node sh
```

Add the following lines to the `node/.config/config.yml`:

```yml
listenGrpcMultiaddr: /ip4/0.0.0.0/tcp/8337                                                                                                                                                    
listenRESTMultiaddr: /ip4/0.0.0.0/tcp/8338    
```

Restart

```sh
docker compose restart
```

You can then run commands against the node, i.e. token balance:

```sh
GOEXPERIMENT=arenas go run ./... -balance
```

---

As an additional note I also made sure to back up my `keys.yml` file but I agree with others on this thread a better approach would be to mount a volume.

```sh
docker cp quilibrium-node-1:/opt/ceremonyclient/node/.config/keys.yml ./
```

CassOnMars (approved):

CassOnMars:
Thanks! This is awesome! Will merge in with next update, trying to avoid thrashing nodes that auto-reload on git changes when folks are still getting caught up.

mscurtescu:
> Thanks! This is awesome! Will merge in with next update, trying to avoid thrashing nodes that auto-reload on git changes when folks are still getting caught up.

Perfect, I'll let you merge at the right time. Will follow up with more PRs related to docker.

mscurtescu:
FYI this followup PR solves the `listenGrpcMultiaddr` and `listenRESTMultiaddr` configuration issue:
https://github.com/QuilibriumNetwork/ceremonyclient/pull/49

PR #56, opened 2024-02-20
add .dockerignore and docker instructions

mscurtescu:
- add `.dockerignore` to trim files copied into the image
- add `DOCKER-README.md` with Docker specific instructions
- update the node restart policy to `unless-stopped`

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -6,7 +6,7 @@ services:
   node:
     build: ./
     image: quilibrium
-    restart: always
+    restart: unless-stopped
     environment:
       - DEFAULT_LISTEN_GRPC_MULTIADDR=/ip4/0.0.0.0/tcp/8337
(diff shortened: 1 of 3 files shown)


docker-compose.yml
AvAcalho, line 1: Glad to see the progress!! I've mixed feeling with `restart: unless-stopped` I'm using `on-failure:5` to mitigate successfully `signal: killed` issue. What you think about it ?

mscurtescu: The problem with on-failure is that the service will not restart after a docker restart (if there's a docker update for example), which also means no service restart on system reboot (kernel update for example). So you end up with extended downtime.

Ideally production should use always. unless-stopped is a good compromise while you are still making frequent changes and tinker.

CassOnMars (approved):

PR #66, opened 2024-02-22
Adding resource management section to Docker readme file

sercanCyberVision:
This PR adds a section to the README detailing how to configure resource limits and reservations for the client within a Docker deployment.

### DIFF
--- a/DOCKER-README.md
+++ b/DOCKER-README.md
@@ -72,3 +72,18 @@ Run the DB console:
 docker compose exec node go run ./... -db-console
 ```
 
+## Resource management (Optional)
+To ensure that your client performs optimally within a specific resource configuration, you can specify resource limits and reservations in the node configuration as illustrated below. This configuration helps in deploying the client with controlled resource usage, such as CPU and memory, to avoid overconsumption of resources in your environment.
+```yaml
+services:
+  node:
+    # Some other configuration sections here
+    deploy:
+      resources:
+        limits:
+          cpus: '4'  # Maximum CPU count that the container can use
+          memory: 16G  # Maximum memory that the container can use
+        reservations:
+          cpus: '2'  # CPU count that the container initially requests
+          memory: 8G  # Memory that the container initially requests
+```


mscurtescu:
Great minds think alike :-)

I just added the resource management instructions to the `docker-compose.yml` file, see https://github.com/QuilibriumNetwork/ceremonyclient/pull/75

We can add an explanation to `DOCKER-README.md`, as your PR does here. Maybe slightly change the working and point to `docker-compose.yml`?

Also, I noticed that with `docker compose up` the reservations are not enforced, for example 12 CPUs might be required and the host has only 6, the container will still be started. Do you know how to actually enforce this?

sercanCyberVision:
@mscurtescu, indeed, it seems like we're on the same wavelength here! :)

I see your point about using default values as the suggested system requirements, which is better idea. I'm happy to close my PR if you'd like to update the `DOCKER-README.md` with your default values along with the explanations from this PR.

Regarding the enforcement of Docker reservations, you're right. It seems Docker alone doesn't enforce these reservations, and it appears an orchestration tool is necessary for actual enforcement. Docker Swarm documentation suggests that it can be done https://docs.docker.com/engine/swarm/services/#reserve-memory-or-cpus-for-a-service.

mscurtescu:
> @mscurtescu, indeed, it seems like we're on the same wavelength here! :)
> 
> I see your point about using default values as the suggested system requirements, which is better idea. I'm happy to close my PR if you'd like to update the `DOCKER-README.md` with your default values along with the explanations from this PR.
> 
> Regarding the enforcement of Docker reservations, you're right. It seems Docker alone doesn't enforce these reservations, and it appears an orchestration tool is necessary for actual enforcement. Docker Swarm documentation suggests that it can be done https://docs.docker.com/engine/swarm/services/#reserve-memory-or-cpus-for-a-service.

Added your resource management section to DOCKER-README.md, thanks a lot.

And thanks for the clarifications around enforcement.

PR #92, opened 2024-03-02
Readme Update

elprofessor-creator:
Updated readme with more detailed installation and upgrade instuctions.


README.md
```diff
@@ -5,6 +5,50 @@ This release, mirrored to GitHub, is the Dawn release, which contains the
...
+Copy and paste the 3 lines below into the file. The values below are six hundred million.
+
+    #Increase buffer sizes for better network performance
+    net.core.rmem_max=600000000
+    net.core.wmem_max=600000000
+
+Save and exit then
+    sudo sysctl -p
+
+Then reboot.
```
Freekers, line 45: Reboot is not needed

Freekers, line 45: Reboot is not needed

README.md
```diff
@@ -29,8 +73,24 @@ The peer id will be printed to stdout.
...
+
+    listenGrpcMultiaddr: /ip4/127.0.0.1/tcp/8337
+    listenRESTMultiaddr: /ip4/127.0.0.1/tcp/8338
+
+Save and exit
+
+Ensure that port 8337 among other neeeded ports are enabled via firewal.
+
+    sudo ufw enable
+    sudo ufw allow 22
```
Freekers, line 88: Port 22 is unrelated to Quilibrium

README.md
```diff
@@ -29,8 +73,24 @@ The peer id will be printed to stdout.
...
+    sudo ufw allow 8336
+    sudo ufw allow 8337
+    sudo ufw allow 8338
```
Freekers, line 91: Why would you open port 8337 and 8338? In line 80 and 81 you restrict them to localhost (127.0.0.1). Additionally, from the README: "Please note: this interface, while read-only, is unauthenticated and not rate- limited. It is recommended that you only enable if you are properly controlling access via firewall or only query via localhost."

Freekers (commented):
Steps are taken from Demipoet's guide, which is fine, but some of them don't make sense.

elprofessor-creator:
> Steps are taken from Demipoet's guide, which is fine, but some of them don't make sense.

Yes, Demipoets guide is working well but many users not able to copy paste commands thats why I would like to add them here to help node runners.

zenodeapp:
Hiya, thank you for this. This somewhat helped me get more reassurance for strolling on the right path!

README.md
```diff
@@ -5,6 +5,49 @@ This release, mirrored to GitHub, is the Dawn release, which contains the
...
+    GOPATH=$HOME/go
+    PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+
+On command line, run 
+    ~/.bashrc
+
+Check GO Version
+    go version
+
+It must show "go version go.1.20.13 linux/amd64"
```
CassOnMars, line 28: These install requirements would install 1.20.14, so it should show `go version go1.20.14 linux/amd64`. That being said, the install instructions are specific to particular Linux distros, so we should probably have separate instructions for each platform if we're going to add these

CassOnMars:
I appreciate the details of the steps, we should get this squared into a README-linux.md or similar

elprofessor-creator:
> I appreciate the details of the steps, we should get this squared into a README-linux.md or similar

Hi @CassOnMars ,
I closed this pr and created new one at https://github.com/QuilibriumNetwork/ceremonyclient/pull/96

Readme is original and README-linux.md is added.

PR #96, opened 2024-03-04
Linux instructions added.

elprofessor-creator:


README-linux.md
```diff
@@ -0,0 +1,242 @@
...
+    GOPATH=$HOME/go
+    PATH=$GOPATH/bin:$GOROOT/bin:$PATH
+
+On command line, run 
+    ~/.bashrc
+
+Check GO Version
+    go version
+
+It must show "go version go.1.20.13 linux/amd64"
```
CassOnMars, line 28: This should be 1.20.14 given the above instructions, but we should probably make sure these instructions direct people to the latest of 1.20

Freekers (commented):
https://github.com/QuilibriumNetwork/ceremonyclient/blob/main/node/poor_mans_cd.sh does exactly this (autoupdate).

CassOnMars (approved):
will merge in with next update, thanks!

PR #102, opened 2024-03-06
multi stage Dockerfile and Taskfile

mscurtescu:
- switching to a multi stage Dockerfile
  - build stage has golang
  - installs the `node` binary
  - installs the `grpcurl` tool
  - the target stage based on [Alpine Linux](https://www.alpinelinux.org/)
  - copies over the `node` and `grpcurl` binaries and a couple of json files
   - much smaller image, 151MB vs 1.18GB
   - safer since neither source code nor `go` are present in production
- adding [Taskfile](https://taskfile.dev/) task runner to help with building and running
- adding `healthcheck` to `docker-compose.yml`
-  `docker-compose.yml` can be customised through `.env`
   - `.env.example` provided as a sample

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,26 @@
-FROM golang:1.20.14-alpine3.19
+FROM golang:1.20.14-alpine3.19 as build
+
+ARG GIT_COMMIT
+
+ENV GOEXPERIMENT=arenas
+
+RUN apk update
+RUN apk add git
+
+WORKDIR /opt/quilibrium
+
+RUN git clone https://github.com/QuilibriumNetwork/ceremonyclient.git
+
+WORKDIR /opt/quilibrium/ceremonyclient
+
+RUN git checkout $GIT_COMMIT
+
+WORKDIR /opt/quilibrium/ceremonyclient/node
+
+RUN go install ./...
+RUN go install github.com/fullstorydev/grpcurl/cmd/grpcurl@latest
+
+FROM alpine:3.19
 
 ARG GIT_COMMIT=unspecified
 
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,23 +1,38 @@
...
-
-WORKDIR /opt/ceremonyclient
-
-COPY . . 
-
-WORKDIR /opt/ceremonyclient/node
+COPY --from=build /go/bin/node /usr/local/bin
+COPY --from=build /go/bin/grpcurl /usr/local/bin
+COPY --from=build /opt/ceremonyclient/node/ceremony.json /root
+COPY --from=build /opt/ceremonyclient/node/retroactive_peers.json /root
```
mscurtescu, line 34: @CassOnMars are there any other files that might be needed at runtime, other than the above two json files?

Dockerfile
```diff
@@ -1,23 +1,38 @@
...
+WORKDIR /opt/ceremonyclient
+
+COPY . .
+
+WORKDIR /opt/ceremonyclient/node
+
+RUN go install ./...
+RUN go install github.com/fullstorydev/grpcurl/cmd/grpcurl@latest
+
+FROM alpine:3.19
```
mscurtescu, line 14: both the build and the runtime stages must be based on the same linux distro, the other easy option is to base them on Debain

@CassOnMars can you see any issues with Alpine in production?

Taskfile.yaml
```diff
@@ -0,0 +1,94 @@
+# https://taskfile.dev
+
+version: '3'
+
+dotenv:
+  - '.env'
+
+vars:
+  VERSION:
+    sh: cat node/main.go | grep "Quilibrium Node - v" |  grep -Eo '[0-9]+\.[0-9]+\.[0-9]+'
```
mscurtescu, line 10: this is extracting the version from `node/main.go`, the version shows up in `db_console.go` as well, and ideally should be returned by `GetNodeInfo` I think.

@CassOnMars I can send a PR to extract the version as a constant in either a new file or one of the existing one. And add it to `GetNodeInfo` as well. Let me know.

Dockerfile
CassOnMars: No, there will be one last addition for 1.5, but I'll be sure to update this file with it

Dockerfile
CassOnMars: Not to my knowledge, every platform has surprising idiosyncracies though. Eventually it'll make sense to build out a rumpkernel for the image

Taskfile.yaml
CassOnMars: Seems reasonable, I'm supportive :+1:

Dockerfile
mscurtescu: thinking of this, not familiar with rumpkernels, but docker uses the host kernel as far as I know so this might not work with docker?

CassOnMars (approved):

PR #109, opened 2024-03-09
Modify the wrong default volumes value to /root/.config

talentbuilder:
When I deploy docker container today, I found volumes default value is not work in docker-compose.xml and node initialization likes a brand new one, so I entered the internal of container and found that node loaded a different path of config which is /opt/ceremonyclient/node/.config. Hope my information can help, thanks.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       retries: 3
       start_period: 1m
     volumes:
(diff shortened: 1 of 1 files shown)


mscurtescu:
Yes, you are absolutely right. This is a regression from the previous structure of the container.

@CassOnMars please approve and merge when you get a chance

PR #124, opened 2024-03-14
fix self test for disk space on a fresh install

shyba:
Hi, :wave: 

I'm not sure what is the process, but I had to fix this line in order to start a fresh install. Otherwise it will try to get disk usage from a folder that doesn't exist yet.
I'm not sure if the correct approach would be to create that folder instead, but here is my attempt at it that worked for my node.


shyba:
was fixed in https://github.com/QuilibriumNetwork/ceremonyclient/pull/125/files#diff-94db04b944e707e22c8149af5679a2703a824073712752f40b21c96acede5bdfR212

PR #128, opened 2024-03-15
Add new bootstrap peers

agostbiro:
Adds two new bare metal bootstrap peers both with high end AMD cpus and 128GB ram. The first one is in the US and the second is in the EU. The first one is at Latitude and the second one is at an untested new provider.


CassOnMars (approved):
will merge in with next update, thanks!

PR #134, opened 2024-03-18
Add commands to check version

Freekers:
Added commands to check local version (while node is running) and latest available version on GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,20 @@ The confirmed token balance will be printed to stdout in QUILs.
 
 Note that this feature requires that [gRPC support](#experimental--grpcrest-support) is enabled.
 
+## Version number check
+
+The version number is displayed during startup of the node. In order to find the version number of a running node, execute the following command from the `node/` folder in a new/seperate terminal window:
+```
+printf "Your local Quilibrium node version is " && printf "$(cat config/version.go | grep -A 1 'func GetVersion() \[\]byte {' | grep -Eo '0x[0-9a-fA-F]+' | xargs printf '%d.%d.%d')" && printf "\n"
+```
+The local node version number will be printed to stdout.
+
+To check the latest node version on GitHub, execute the following command:
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,6 +46,20 @@ The confirmed token balance will be printed to stdout in QUILs.
 
 Note that this feature requires that [gRPC support](#experimental--grpcrest-support) is enabled.
 
+## Version number check
+
+The version number is displayed during startup of the node. In order to find the version number of a running node, execute the following command from the `node/` folder in a new/seperate terminal window:
+```
+printf "Your local Quilibrium node version is " && printf "$(cat config/version.go | grep -A 1 'func GetVersion() \[\]byte {' | grep -Eo '0x[0-9a-fA-F]+' | xargs printf '%d.%d.%d')" && printf "\n"
```
mscurtescu, line 53: this assumes that the source code is available in production, which ideally should not be the case

if the source code is present then you can use the existing `Taskfile.yaml` and run `task status` which does a similar thing.

I think a better solution is to add the version to either `GetNodeInfo`, but that would return a base64 encoded string, and/or add a `-node-info` command (similar to `-balance`). I am looking into implementing these two.

Running the `-db-console` command currently will also show you the version, but it starts full screen and you have to exit.

Freekers: Agreed, but unfortunately I'm not a developer so this was the best I could do. The question of 'which version am I running' or 'which version is latest' is asked quite often in Telegram and Discord, so this might be a nice intermediate solution.

mscurtescu: the which version question is answered by `-db-console`

the PR that adds version to GetNodeInfo is under construction, coming soon

that being said, I am only offering my opinion, I am not blocking anything

Freekers: Would it also be possible to include the latest available version on GitHub in the GetNodeInfo command or create a separate command for that? Because knowing the version of your local node is just half the story imho.

mscurtescu:
This PT adds a `-node-info` command/flag that also shows the version:
https://github.com/QuilibriumNetwork/ceremonyclient/pull/141

Freekers:
Closing due to https://github.com/QuilibriumNetwork/ceremonyclient/pull/141

PR #136, opened 2024-03-19
bandwidth test: pause between retries

shyba:
Most of the peers that lack a bandwidth test are new to pubSub, specially during start up, so they lack a multiaddr at this point.

The current code keeps retrying them without a pause, which showed up on profiler due high CPU usage.

The following change adds a random pause to make those missing an address wait before retry, although I'm not sure for how long they should be retried (I can open a follow up PR that handles pending peers better, if that is useful)

### DIFF
--- a/node/consensus/master/master_clock_consensus_engine.go
+++ b/node/consensus/master/master_clock_consensus_engine.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"io"
 	"math/big"
+	random "math/rand"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


CassOnMars:
Nice catch! Randomness is a bit of a tough call – it would work better for CPU compared to current case, but it's still possible some peers may never pan out (misbehavior), or may take a little longer if the mesh is going through heavy thrashing between updates. I'm going to hook it into the address stream from libp2p and make it wait on a viable address so no re-attempts/randomness is required and just fail it out after some time, where subsequent announcements will make it through. Will get it rolled into 1.4.11. 🙏🏻

shyba:
Closing as it will be handled in 1.4.11

Thank you!

PR #138, opened 2024-03-21
p2p suggestions: save bootstrap file

shyba:
I have some suggestions for the p2p code, so instead of just describing them I decided to open a PR to show what I mean. Feel free to adapt parts of it as/if it feels right.

* Save and restore top 100 connected peers as a bootstrap file to aid on restarts. This may later be used on the `reconnect` function to reboostrap in case of a connection loss. I often see this feature across p2p clients.

EDIT: removed other changes that were solved with peer info manager and the last qol

This PR now only adds support for a bootstrap file (should help bringing connectivity to a new node from an existing one or during reboots, but wont change anything on already good ones)

### DIFF
--- a/node/main.go
+++ b/node/main.go
@@ -16,6 +16,7 @@ import (
 	"runtime"
 	"runtime/pprof"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


shyba:
closing as it doesn't seem to make sense anymore. It is easy to rejoin the network with the new bootstrap nodes

PR #148, opened 2024-03-27
simplify checks and avoid loop on collect

shyba:
If the max frame from the best peer found is latest+1 it will loop for a bit. That is the second else for the inner if that is missing. This simplifies the logic to always break or sleep.


shyba:
already fixed in https://github.com/QuilibriumNetwork/ceremonyclient/commit/52a197368e3b5efaea9a43c52cbe53868ae04641#diff-d24942f25e44377c3b91dedd51ba785ffe4c103ee4ba4894bcec81c78ee1ad6bL277

PR #149, opened 2024-03-27
Master sync: parallelism, fix off by one, ignore far frames, separate bandwidth test in another thread and log progress

shyba:
Master sync happens silently for new nodes, so I was confused thinking it was stuck. This PR makes it log master frame events to info so we can see it happening.

Also, I realized bandwidth tests can arrive in waves, sometimes with dead/slow peers, stalling the master sync validation for a while. I separated a thread for it, although it still feels like it could use more CPU for new nodes.

This is a new part of the code I'm still understanding, so sorry if I got confused. Those are mostly suggestions, feel free to adapt or close.

### DIFF
--- a/node/consensus/master/consensus_frames.go
+++ b/node/consensus/master/consensus_frames.go
@@ -44,7 +44,7 @@ func (e *MasterClockConsensusEngine) GetMostAheadPeers() (
 	max := frame.FrameNumber + 10
 
 	var peers [][]byte = [][]byte{}
-	peerMap := e.peerInfoManager.GetPeerMap()
+	peerMap := e.peerInfoManager.GetPeersBySpeed()
 	for peerId, v := range peerMap {
 		if v.MasterHeadFrame > max {
 			peers = append(peers, []byte(peerId))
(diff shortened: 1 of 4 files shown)


shyba:
added:
* fix off by one in pending process (it was deleting the head+1 after advancing the head, so head+2)
* process pending even when the current one is not pending (may have closed a cycle to a long chain)
* ignore far frames, they get in the way of far behind nodes using CPU and memory
* start as many validators as there are CPU cores
* on master sync, check first frame of every message but then send them to the parallel verifier

0xOzgur:
I applied your code to my 0 frame nodes and it is nice to see that actually recieving frames.
Also, as you mentioned that it makes higher CPU usage.

CassOnMars:
appreciate the off-by-one catch! The sorting by speed matters less for the master frame sync, and it's necessary to keep cpu burn low on processing master frames otherwise it'll potentially block out valuable core count for handling data frames/application calls, which is why it's not a max-out on CPU. I did notice there potentially exists a condition where the off-by-one change will fail, and have addressed this in the next update.

PR #157, opened 2024-04-04
allows pinning version in poor_mans_cd.sh

bayoumymac:
## what

currently when running`poor_mans_cd.sh`, there is not a convenient way to rollback a version while continuing to use the safety benefits that comes with the script

this diff adds a new env var `PINNED_VERSION` that when present uses `git reset` to rollback to a specific hash and doesn't check for future updates

if the env var isn't set, script operates as usual

### DIFF
--- a/node/poor_mans_cd.sh
+++ b/node/poor_mans_cd.sh
@@ -35,6 +35,12 @@ kill_process() {
 
 kill_process
 
+if [ $PINNED_VERSION != "" ]; then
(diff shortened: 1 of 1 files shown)


rak16:
Looks like a fine short term solution so that people can continue using `poor_mans_cd.sh` but a relatively longer term solution would be to make the script pull only the signed releases – which shouldn't be that hard to do, would it? 🤔

CassOnMars:
Closing as release_autorun.sh in the release branch now does strict versioning

PR #167, opened 2024-04-23
Liberty

CassOnMars:
And what country can preserve its liberties if their rulers are not warned from time to time that their people preserve the spirit of resistance?


a372115389:
And what country can preserve its liberties if their rulers are not warned from time to time that their people preserve the spirit of resistance?

Issue #175, opened 2024-05-12
Docker qclient cross-mint error


0xOzgur:
Normally docker node commands has structure like below

```
docker compose exec node qclient help
docker compose exec node qclient token help
docker compose exec node qclient token balance
```

When using crossmint function with this structure, throws error
```
`docker compose exec node qclient cross-mint 0x0000000`
config directory doesn't exist: ../node/.config/
```

We need to run command with config flag.
`docker compose exec node qclient cross-mint --config .config 0x0000000`

@agostbiro @mscurtescu

agostbiro:
Thanks for the detailed report @0xOzgur ! I'm not familiar with the Docker setup, so I'd defer to @mscurtescu

mscurtescu:
Up until 1.4.17 the `qclient` command was just a dummy placeholder, no real config was loaded, the `cross-min` subcommand is the first that is also loading the config.

The config folder is hard coded to load from a relative folder for `qclient`, and that will not work as expected if the binary is installed and run from PATH:
https://github.com/QuilibriumNetwork/ceremonyclient/blob/main/client/cmd/root.go#L28

Working on a PR that fixes this...

mscurtescu:
https://github.com/QuilibriumNetwork/ceremonyclient/pull/176

0xOzgur:
@mscurtescu Great job, thank you!

PR #176, opened 2024-05-13
shift default config directory of qclient under current folder

mscurtescu:
`qclient` by default expects the config folder under `../node/.config/` which is not working when the binary is run from PATH.

The workaround is to run the command like:
```shell
qclient <sub-command> --config .config <args>
```

This PR fixes the config loading to default it to `.config/` (that is `.config/` under current folder).

Related issue: https://github.com/QuilibriumNetwork/ceremonyclient/issues/175


mscurtescu:
Tested locally, works as expected.

PR #186, opened 2024-05-20
Don't run self-test in DHT-only mode

agostbiro:
## Problem

The node tries to create a self-test when ran with the `-dht-only` flag, but it doesn't load the KZG ceremony data in DHT-only mode which leads to a crash.

## Solution

Don't run self-test when the `-dht-only` flag is set.

## Testing

I tested by starting a node locally in default mode with and without existing self-test and with the `-dht-only` flag.


CassOnMars:
Great catch, thanks! 🙏

PR #187, opened 2024-05-20
Change bootstrap servers to DHT-only peers

agostbiro:
Changing my bootstrap servers to DHT-only peers with somewhat lower specs. One of the new ones is in the US and the other one is in Switzerland. Both use reliable providers and have 10Gbps network interfaces.


CassOnMars:
🙏🏻 thanks! can you re-target to branch v1.4.18?

PR #211, opened 2024-05-26
ReleaseSolves an issue where arm machines cannot execute scripts

dreamking02:
Add the arch variable to get whether the version of the system is amd or arm


agostbiro:
Thanks for the fix @dreamking02 ! Could you base your PR branch on the `main` branch in this repo to avoid other changes in the PR? Also, could you please add a new line at the end of the script?

PR #212, opened 2024-05-26
fix get node info maxFrame return

0xluk:
switched get node info response to use masterClock frame for maxFrame field


CassOnMars:
Thanks!

PR #213, opened 2024-05-26
Update main.go to fix Q logo in cli

rak16:
Q logo is not appearing correctly on the terminal while running node. Printing a new line character after "Signature check passed" will fix it


CassOnMars:
Appreciate this. Can you change the target to v1.4.18-patch-1? We can roll it into our patch update.

PR #224, opened 2024-05-27
fix arm64 architecture error

feeeei:
There is no `$arch` variable to identify the architecture, `uname -m` should be used to determine the architecture.


CassOnMars:
what distro are you running into this issue on?

some arm64 processors give totally different uname responses which is why we avoided it

feeeei:
None of my Linux devices have the `$arch` variable. 

I have these OS:
- Ubuntu 22.04.3 LTS
- Ubuntu 22.04.4 LTS
- Ubuntu 24.04 LTS
- MacOS 14.5

Would it be more universal to use `uname`?

0xOzgur:
@Freekers sound like you are right.
I wasn't able to run script at my ubuntu 24.04 LTS arm machines. I chanched to uname then works.

VasylTretiakov:
Should this fix be combined with #211  ?

PR #230, opened 2024-05-29
v1.4.18-patch-2

CassOnMars:
## Changelog
- Switch to using IPC model for data workers to alleviate GC pressure bug
- Fix proof serialization bug in wesolowski

Builds produced via a MacOS based builder.

## Signer steps
Commands to produce digests:
```
openssl sha3-256 -out node-1.4.18-darwin-arm64.dgst node-1.4.18-darwin-arm64
openssl sha3-256 -out node-1.4.18-linux-amd64.dgst node-1.4.18-linux-amd64
openssl sha3-256 -out node-1.4.18-linux-arm64.dgst node-1.4.18-linux-arm64
```

Signatories, verify above, then produce signatures:
```
openssl pkeyutl -sign -inkey <path-to-key> -rawin -in node-1.4.18-darwin-arm64.dgst -out node-1.4.18-darwin-arm64.dgst.sig.<key-id>
openssl pkeyutl -sign -inkey <path-to-key> -rawin -in node-1.4.18-linux-amd64.dgst -out node-1.4.18-linux-amd64.dgst.sig.<key-id>
openssl pkeyutl -sign -inkey <path-to-key> -rawin -in node-1.4.18-linux-arm64.dgst -out node-1.4.18-linux-arm64.dgst.sig.<key-id>
```

Add to PR via pull request against v1.4.18-patch-2


VasylTretiakov:
Should not we update the [Minimum System Requirements](README.MD) to reflect the minimum virtual core count of 4 (now that the app will panic/stop when it sees fewer cores)?
[Linux README](README-linux.MD) also looks outdated, but, if no one actively uses it, it could be dealt with (possibly removed if instructions get published separately) later.

PR #248, opened 2024-06-21
Update release_autorun.sh so kill_process is version agnostic

peterjameson:


CassOnMars (approved):
awesome, thank you for catching this!

PR #249, opened 2024-06-25
release_autorun improvement 

nero08:
Separate managment of autoupdate and crash detector to avoid waiting too much time when node crash

### DIFF
--- a/node/release_autorun.sh
+++ b/node/release_autorun.sh
@@ -47,22 +47,17 @@ fetch() {
             new_release=true
         fi
     done
-    echo $new_release
 }
 
 git_update_manager() {
     while true; do
-        new_release=$(fetch)
-        git fetch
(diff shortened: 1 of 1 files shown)


node/release_autorun.sh
```diff
@@ -58,27 +47,49 @@ fetch() {
...
+    echo $new_release
 }
 
-fetch
-
-kill_process
+git_update_manager() {
+    while true; do
+        new_release=$(fetch)
+        git fetch
```
CassOnMars, line 56: I appreciate the work that went into this PR, but we cannot restore periodic git fetching, that is a firm condition on GitHub's willingness to restore this repository.

CassOnMars (changes requested):

nero08:
Hello, I did a change to execute git pull only if new release available

PR #252, opened 2024-06-29
added sig.6 files

demipoet:
Added signer #6 .sig files


CassOnMars:
added to CDN:
https://releases.quilibrium.com/node-1.4.20.1-darwin-arm64.dgst.sig.6
https://releases.quilibrium.com/node-1.4.20.1-linux-arm64.dgst.sig.6
https://releases.quilibrium.com/node-1.4.20.1-linux-amd64.dgst.sig.6

PR #289, opened 2024-09-01
Remove extra difficulty check for proofs

tjsturos:
There was a redundant check for if the increment was over 800k. 

This would be redundant due to the first check of "if the difficulty was under 25k" would be met at anything over 700k.

Moved to a function, due to it being used twice (in the Calculating and Verifying of the Proof) and to be able to add tests.

### DIFF
--- a/node/crypto/wesolowski_frame_prover.go
+++ b/node/crypto/wesolowski_frame_prover.go
@@ -582,13 +582,11 @@ func (w *WesolowskiFrameProver) VerifyWeakRecursiveProof(
 func (w *WesolowskiFrameProver) CalculateChallengeProofDifficulty(
 	increment uint32,
 ) uint32 {
-	difficulty := 200000 - (increment / 4)
-
-	if difficulty < 25000 {
-		difficulty = 25000
+	if increment >= 700000 {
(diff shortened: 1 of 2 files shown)


CassOnMars (approved):

node/crypto/wesolowski_frame_prover.go
```diff
@@ -579,15 +579,24 @@ func (w *WesolowskiFrameProver) VerifyWeakRecursiveProof(
 	}
 }
 
+func (w *WesolowskiFrameProver) CalculateChallengeProofDifficulty(
+	increment uint32,
+) uint32 {
+	difficulty := 200000 - (increment / 4)
```
CassOnMars, line 585: the increment check is needed because this is a uint32, it'll underflow

CassOnMars:
looks good minus the underflow check

node/crypto/wesolowski_frame_prover.go
tjsturos: Reordered the function to only do the calculation on increments less than 700k

CassOnMars (approved):
awesome, thanks!

PR #302, opened 2024-10-16
fix get proof

SourceLink:


node/execution/intrinsics/token/token_execution_engine.go
```diff
@@ -321,14 +321,14 @@ func NewTokenExecutionEngine(
 				if err == nil {
 					for _, proof := range proofs {
 						if proof.IndexProof != nil && len(proof.IndexProof) != 0 {
-							if proof.Index < inc {
```
CassOnMars, line 324: this is correct. Difficulty is not the right value to use here

PR #315, opened 2024-10-29
Ensure direct peers are not pruned by resource limits

petricadaipegsp:
This PR makes both the bootstrap peers and the direct peers be part of the resource limiter allowed list. This ensures that the connections are not pruned by the resource manager.

cc @CassOnMars


CassOnMars:
I agree with this addition, but it is worth noting that unless low/high watermarks are explicitly configured, the resource manager here never gets set, instead using the default one

PR #319, opened 2024-10-31
Fix peer discovery

petricadaipegsp:
This PR fixes an issue introduced by commit 7eaf8e05412086ed5f52023f7211aea487865afc where the peer lookup discovery stops early instead of skipping peers. The symptoms of this are that peers are not discovered (duh), and also over time there is a CPU and memory leak due to the background lookup processes.

It also reintroduces the parallel peer connections, but compared to the old parallel connection method, it waits for the connection attempts to end before finishing the discovery. This ensures that multiple discovery processes don't overlap (well, minus the initial one, but one time is fine).

### DIFF
--- a/node/p2p/blossomsub.go
+++ b/node/p2p/blossomsub.go
@@ -924,9 +924,10 @@ func discoverPeers(
 	routingDiscovery *routing.RoutingDiscovery,
 	init bool,
 ) {
-	logger.Info("initiating peer discovery")
-
 	discover := func() {
+		logger.Info("initiating peer discovery")
+		defer logger.Info("completed peer discovery")
+
 		peerChan, err := routingDiscovery.FindPeers(
 			ctx,
 			getNetworkNamespace(p2pConfig.Network),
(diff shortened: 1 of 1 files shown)


node/p2p/blossomsub.go
```diff
@@ -936,28 +937,33 @@ func discoverPeers(
...
 		}
 
+		wg := &sync.WaitGroup{}
+		defer wg.Wait()
 		for peer := range peerChan {
 			peer := peer
-			if peer.ID == h.ID() ||
-				h.Network().Connectedness(peer.ID) == network.Connected ||
-				h.Network().Connectedness(peer.ID) == network.Limited {
-				return
```
petricadaipegsp, line 944: The `return` here basically stops the lookup process virtually instantly, but it does not correctly stop the `FindPeers` background goroutine, and that causes a leak.

CassOnMars: Nice catch, thanks!

CassOnMars (approved):

PR #320, opened 2024-10-31
Additional peering connection improvements

petricadaipegsp:
This PR proposes two changes to the peering mechanism:
1. When considering if a bitmask needs more peers, do not take into account the bootstrap and the direct nodes. This achieves two things - it reduces the dependence on the bootstrap for non bootstrap operations (as the bootstraps are overloaded), and it ensures that a direct peers mesh is not accidentally solipsistic - well connected within itself, but not with the rest of the network.
2. Ensure that at least one bootstrap peer is always connected, in order to allow peer lookups, should they be required. Additional concurrency guards have been added - `reconnect` won't finish until it has either connected or at least attempted to connect to the bootstraps, so there won't be a goroutine leak.

### DIFF
--- a/node/p2p/blossomsub.go
+++ b/node/p2p/blossomsub.go
@@ -384,7 +384,7 @@ func NewBlossomSub(
 	bs.h = h
 	bs.signKey = privKey
 
-	allowedPeerIDs := map[peer.ID]struct{}{}
+	allowedPeerIDs := make(map[peer.ID]struct{}, len(allowedPeers))
 	for _, peerInfo := range allowedPeers {
 		allowedPeerIDs[peerInfo.ID] = struct{}{}
 	}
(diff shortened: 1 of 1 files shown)


node/p2p/blossomsub.go
```diff
@@ -718,12 +733,20 @@ func initDHT(
...
+	}
 	go func() {
+	outer:
 		for {
 			time.Sleep(30 * time.Second)
-			if len(h.Network().Peers()) == 0 {
-				reconnect()
+			for _, p := range h.Network().Peers() {
+				if _, ok := bootstrapPeerIDs[p]; ok {
+					continue outer
```
CassOnMars: this looks like it'll loop until it's disconnected from bootstraps?

petricadaipegsp: The sleep on L743 still applies - it won't spin. So the loop starts with a sleep, it looks at the current peers, and if a bootstrapper is found, it jumps back to the sleep. otherwise it goes into `reconnect`.

petricadaipegsp: Simplified without label jumping.

PR #327, opened 2024-11-02
On demand bootstrap reconnection

petricadaipegsp:
It's a day that ends in `y`, so we're fixing peering. Today's fixes include:
- Lower the ping timeout to 20 seconds. No connections observed in the wild responded in more than 20 seconds to pings during testing.
- Wait for the connection to be identified after establishing it, in order to increase the odds that it is a fully opened connection.
- Reconnect to bootstrap servers before attempting to do a peer lookup, if we are not connected to at least 3 bootstrap servers.
- Do not protect bootstrap peers. We'll reconnect on demand instead.
- Do not attempt to reconnect to all bootstrap servers, and instead attempt to connect to at least 3 servers. The order of the connections is randomized in order to avoid concentrating load. Every reconnection attempt has another connection order.

### DIFF
--- a/node/p2p/blossomsub.go
+++ b/node/p2p/blossomsub.go
@@ -15,7 +15,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/libp2p/go-libp2p"
@@ -26,14 +25,12 @@ import (
 	"github.com/libp2p/go-libp2p/core/host"
 	"github.com/libp2p/go-libp2p/core/network"
 	"github.com/libp2p/go-libp2p/core/peer"
-	"github.com/libp2p/go-libp2p/core/peerstore"
 	"github.com/libp2p/go-libp2p/core/protocol"
 	"github.com/libp2p/go-libp2p/p2p/discovery/routing"
 	"github.com/libp2p/go-libp2p/p2p/discovery/util"
 	rcmgr "github.com/libp2p/go-libp2p/p2p/host/resource-manager"
 	"github.com/libp2p/go-libp2p/p2p/net/connmgr"
 	"github.com/libp2p/go-libp2p/p2p/net/swarm"
-	"github.com/libp2p/go-libp2p/p2p/protocol/identify"
 	"github.com/libp2p/go-libp2p/p2p/protocol/ping"
 	"github.com/mr-tron/base58"
 	ma "github.com/multiformats/go-multiaddr"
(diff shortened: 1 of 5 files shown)


node/p2p/blossomsub.go
```diff
@@ -144,20 +149,13 @@ func NewBlossomSubStreamer(
...
+	_ = initDHT(
 		ctx,
-		p2pConfig,
 		logger,
 		h,
 		false,
 		bootstrappers,
 	)
-	routingDiscovery := routing.NewRoutingDiscovery(kademliaDHT)
-	util.Advertise(ctx, routingDiscovery, getNetworkNamespace(p2pConfig.Network))
```
petricadaipegsp, line 156: Advertising while being a `SubStreamer` is not required, as the peer is not expected to participate in the mesh.

node/p2p/blossomsub.go
```diff
@@ -602,7 +600,7 @@ func (b *BlossomSub) GetRandomPeer(bitmask []byte) ([]byte, error) {
 // up to 3 times to ensure they are still reachable. If the peer is not reachable after
 // 3 attempts, the connections to the peer are closed.
 func monitorPeers(ctx context.Context, logger *zap.Logger, h host.Host) {
-	const timeout, period, attempts = time.Minute, time.Minute, 3
+	const timeout, period, attempts = time.Minute, 20 * time.Second, 3
```
petricadaipegsp, line 643: We're lowering the ping timeout in order to catch the half open connections faster.

node/p2p/blossomsub.go
```diff
@@ -646,21 +644,12 @@ func monitorPeers(ctx context.Context, logger *zap.Logger, h host.Host) {
 		case <-ctx.Done():
 			return
 		case <-time.After(period):
-			// This is once again a snapshot of the peers at the time of the ping. If new peers
+			// This is once again a snapshot of the connected peers at the time of the ping. If new peers
 			// are added between the snapshot and the ping, they will be pinged in the next iteration.
 			peers := h.Network().Peers()
-			connected := make([]peer.ID, 0, len(peers))
```
petricadaipegsp, line 652: `Peers()` already returns connections which are the connected or limited state, so the filtering was superfluous.

node/p2p/blossomsub.go
```diff
@@ -738,17 +736,20 @@ func initDHT(
...
 	}
 	go func() {
+		const minBootstrapPeers = 3
 		for {
 			time.Sleep(30 * time.Second)
-			found := false
+			count := 0
 			for _, p := range h.Network().Peers() {
 				if _, ok := bootstrapPeerIDs[p]; ok {
-					found = true
```
petricadaipegsp, line 746: The beacon was going to be 'the one bootstrap peer' that we would be connected to quite often, and this meant that we wouldn't attempt to reconnect to the bootstrappers if we had at least the beacon. But the beacon is not a DHT server, so we actually had no capability to find peers. This resulted in these peer lookups that finished instantly, with no results.

As such, we now require at least 3 bootstrap peers.

node/p2p/blossomsub.go
```diff
@@ -704,28 +692,38 @@ func initDHT(
...
-					h.Network().Connectedness(peerinfo.ID) == network.Limited {
+				if p.ID == h.ID() ||
+					h.Network().Connectedness(p.ID) == network.Connected ||
+					h.Network().Connectedness(p.ID) == network.Limited {
+					logger.Debug("peer already connected")
 					return
 				}
 
-				if err := h.Connect(ctx, peerinfo); err != nil {
+				h.Peerstore().AddAddrs(p.ID, p.Addrs, peerstore.AddressTTL)
```
petricadaipegsp, line 712: We use `DialPeer` instead of `Connect` because it actually gives us the newly dialed connection, and we can run a health check on it via the `identify` service. This ensures that the connection is not 'connected but not really'.

petricadaipegsp:
Since I cannot make PR chains show up here, here are some upcoming PRs once this one is accepted and merged:
- https://github.com/petricadaipegsp/ceremonyclient/pull/1
- https://github.com/petricadaipegsp/ceremonyclient/pull/2

The second one is particularly interesting as it enables the node to connect to peers whose ID has been learned via PubSub peer list announcements, but for which we don't have a connection address yet. Currently attempting to sync with such peers fails instantly, as we have no address for them.

PR #328, opened 2024-11-04
Trigger automatic peer discovery on frame stall

petricadaipegsp:
References https://github.com/QuilibriumNetwork/ceremonyclient/pull/327

This PR automatically triggers a peer discovery and connection round when the tape head has been stalling for at least 2 minutes. It will then follow an exponential pattern (2 minutes -> 4 minutes -> 8 minutes) of waiting then triggering a new discovery round. Jitter has been added to this wait period in order to avoid hammering the DHT servers excessively.

This mechanism does not replace the PubSub one of ensuring that at least 4 peers are available per bitmask, but instead is meant to bridge the meshes when a partition occurs during steady state. 

It is also slightly slow due to the exponential backoff. We expect that the whole disconnected mesh will run this procedure in order to connect to the remaining mesh - it is not just a node which works towards this, it is the whole disconnected sub-part of the mesh.

### DIFF
--- a/node/consensus/data/data_clock_consensus_engine.go
+++ b/node/consensus/data/data_clock_consensus_engine.go
@@ -7,9 +7,11 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math/big"
+	"math/rand"
 	"sync"
 	"time"
 
+	"github.com/libp2p/go-libp2p/p2p/discovery/backoff"
 	"github.com/multiformats/go-multiaddr"
 	mn "github.com/multiformats/go-multiaddr/net"
 	"github.com/pkg/errors"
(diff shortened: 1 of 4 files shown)


node/consensus/data/data_clock_consensus_engine.go
```diff
@@ -338,6 +340,39 @@ func (e *DataClockConsensusEngine) Start() <-chan error {
 
 	e.state = consensus.EngineStateCollecting
 
+	go func() {
```
petricadaipegsp, line 343: Are there other places where this could've been placed in ? Right now it affects only the data clock consensus engine, but maybe it should check for stalls in all engines long term ?

CassOnMars (approved):

PR #333, opened 2024-11-05
Use peer address while checking if date time proofs should be created

petricadaipegsp:
This PR fixes two checks:
- It uses the peer address instead of the proving address to check ring membership.
- It uses gRPC status codes to propagate not found errors.

### DIFF
--- a/node/consensus/data/data_clock_consensus_engine.go
+++ b/node/consensus/data/data_clock_consensus_engine.go
@@ -20,7 +20,9 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/credentials/insecure"
+	"google.golang.org/grpc/status"
 	"google.golang.org/protobuf/types/known/anypb"
 	"source.quilibrium.com/quilibrium/monorepo/go-libp2p-blossomsub/pb"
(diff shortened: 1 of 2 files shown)


node/rpc/data_worker_ipc_server.go
```diff
@@ -152,10 +153,7 @@ func (r *DataWorkerIPCServer) CalculateChallengeProof(
 	}
 
 	if !found {
-		return nil, errors.Wrap(
```
petricadaipegsp, line 155: Wrapped errors are unfortunately just propagated as strings via gRPC, so the original structure is lost. Status codes however are the standard way of communicating errors.

node/consensus/data/data_clock_consensus_engine.go
```diff
@@ -551,8 +558,8 @@ func (e *DataClockConsensusEngine) Start() <-chan error {
 			frame = nextFrame
 
 			for i, trie := range e.GetFrameProverTries()[1:] {
-				if trie.Contains(e.provingKeyAddress) {
-					e.logger.Info("creating data shard ring proof", zap.Int("ring", i-1))
+				if trie.Contains(peerProvingKeyAddress) {
```
petricadaipegsp, line 561: Is this the last place or do we need to edit others ?

CassOnMars: last place

PR #337, opened 2024-11-06
Sample peers which are ahead during sync

petricadaipegsp:
This PR brings the following changes to the frame syncing process:
- We now look at the _peers_ head instead of the _peer_ ahead. This ensures that we don't re-loop the peer map after every failure.
- We now no longer jump at the highest peer and try to sync with him, but instead look at all of the peers which are ahead of us, and we pick one of them randomly. Peers which are further ahead (frame number wise) are more likely to be tried first before the ones which are closer.

### DIFF
--- a/node/consensus/data/consensus_frames.go
+++ b/node/consensus/data/consensus_frames.go
@@ -8,13 +8,13 @@ import (
 	"golang.org/x/crypto/sha3"
 	"source.quilibrium.com/quilibrium/monorepo/node/config"
 	"source.quilibrium.com/quilibrium/monorepo/node/consensus"
+	"source.quilibrium.com/quilibrium/monorepo/node/consensus/data/internal"
 
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
 	"google.golang.org/grpc"
 	"google.golang.org/protobuf/proto"
 	"source.quilibrium.com/quilibrium/monorepo/node/execution/intrinsics/token/application"
-	"source.quilibrium.com/quilibrium/monorepo/node/p2p"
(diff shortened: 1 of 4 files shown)


node/consensus/data/consensus_frames.go
```diff
@@ -23,29 +23,25 @@ func (e *DataClockConsensusEngine) collect(
 ) (*protobufs.ClockFrame, error) {
 	e.logger.Info("collecting vdf proofs")
 
+	e.syncingStatus = SyncStatusSynchronizing
```
CassOnMars: This was what I was referring to – the sync status shouldn't be set unless sync is being invoked, due to the possibility the node could be an active frame prover (obviously not at the moment but after pre2.0 proofs are done and we can unshackle ourselves from the beacon model) – synchronizing state will halt processing of other messages. I think the approach is awesome, this is the only blocking item.

petricadaipegsp: Should be fixed now.

node/consensus/data/consensus_frames.go
```diff
@@ -24,30 +24,24 @@ func (e *DataClockConsensusEngine) collect(
 	e.logger.Info("collecting vdf proofs")
 
 	latest := enqueuedFrame
-
 	for {
-		peerId, maxFrame, err := e.GetMostAheadPeer(e.latestFrameReceived)
```
petricadaipegsp, line 29: Getting the most ahead peer based only on `e.latestFrameReceived` is actually wrong at times - `latest` may be ahead of `e.latestFrameReceived` now, and we don't need to sync actually.

PR #340, opened 2024-11-08
Sort proofs by frame number while establishing the merkle root

petricadaipegsp:
_This fix is based on a bogus understanding of the proof store and still in testing._


petricadaipegsp:
I don't think there is a bug here, or at least I cannot isolate it. There is at most one proof at a time in the store, and even if the previous root is incorrect, the new proof is stored. Previous claims in this PR (that the previous root should be `nil`ified, or that the proofs in the store should be sorted upon retrieval) are unsupported by the actual code.

PR #346, opened 2024-11-11
Add message validators

petricadaipegsp:
This PR adds a validation at `blossomsub` level for the messages which are routed. It automatically will drop old messages (frames older than an hour, and data peer list announcements older than 10 minutes).

### DIFF
--- a/node/consensus/data/message_validators.go
+++ b/node/consensus/data/message_validators.go
@@ -1,0 +1,101 @@
+package data
+
+import (
+	"time"
+
+	"github.com/libp2p/go-libp2p/core/peer"
+	"go.uber.org/zap"
+	"google.golang.org/protobuf/proto"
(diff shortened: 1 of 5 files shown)


node/consensus/data/message_validators.go
```diff
@@ -0,0 +1,99 @@
...
+		return -1
+	}
+	switch a.TypeUrl {
+	case protobufs.TokenRequestType:
+		tx := &protobufs.TokenRequest{}
+		if err := proto.Unmarshal(a.Value, tx); err != nil {
+			e.logger.Debug("could not unmarshal token request", zap.Error(err))
+			return -1
+		}
+		return 0
```
petricadaipegsp, line 61: There is no timestamp associated with a token request. I am not sure if there should be one, but just FYI.

CassOnMars (approved):

PR #354, opened 2024-11-15
Add transaction timestamps

petricadaipegsp:
This PR adds a timestamp to the token requests and adds validators which automatically drop old transactions. Existing transactions passed around without a timestamp are currently still allowed to fly for backwards compatibility reasons.

It also adds validation of the mint requests, and drops mint requests which are too old when compared to the underlying data time reel.


node/consensus/data/message_validators.go
```diff
@@ -59,7 +59,16 @@ func (e *DataClockConsensusEngine) validateTxMessage(peerID peer.ID, message *pb
 			e.logger.Debug("could not unmarshal token request", zap.Error(err))
 			return p2p.ValidationResultReject
 		}
-		// NOTE: There are no timestamps to be validated for token requests.
+		if tx.Timestamp == 0 {
+			// NOTE: The timestamp was added in later versions of the protocol,
```
petricadaipegsp, line 63: We also _could_ turn this into a breaking change for the protocol in order to cull all of the lingering messages which are not timed out.

Or we make it such that this little escape hatch expires after a particular calendar date (12 hours from release ?) and then messages without a timestamp are automatically ignored.

PR #356, opened 2024-11-16
Add sync timeout and make it async

petricadaipegsp:
This PR sets a fixed timeout (2 seconds per frame) for `sync` requests towards other peers. While the timeout is quite aggressive, it is meant to speed up the sync process by cutting the time spent on unreachable peers.

It also makes the synchronization asynchronous, by moving it to a dedicated goroutine. The sync may be sporadically triggered, but it won't block the potential arrival of new frames from the PubSub (via the time reel itself).


petricadaipegsp:
After https://github.com/QuilibriumNetwork/ceremonyclient/pull/352 is merged, this PR will be refactored to add a dedicated setting for the timeout. In this way we don't need a new release to tune out the timeout.

PR #357, opened 2024-11-16
Use buffered channels in the engines

petricadaipegsp:
This PR proposes that certain unbuffered channels in the engines and the time reels are upgraded to minimally buffered channels. See the comments below on why this may be helpful.


node/consensus/time/data_time_reel.go
```diff
@@ -685,15 +685,16 @@ func (d *DataTimeReel) setHead(frame *protobufs.ClockFrame, distance *big.Int) e
...
+		go func() {
+			select {
+			case d.newFrameCh <- frame:
+			default:
+			}
+		}()
 	}
-	go func() {
-		select {
-		case d.newFrameCh <- frame:
```
petricadaipegsp, line 693: So the trouble is that while we are doing a sync, a new frame may arrive via the PubSub. We can process already that frame while the sync is running. However, since the `newFrameCh` is unbuffered, we won't actually be able to write the new frame (which we are at the same time searching via sync), and we 'lose' this frame that we now have, but didn't process yet. Because of this, we need to wait 20 seconds until the next frame processing interval is kicked again and this frame that we already have is processed.

By making the channel buffered, we will immediately be able to pickup the frame after the sync, and avoid the 20 second delay.

PR #363, opened 2024-11-19
Adaptive reserved cores

petricadaipegsp:
This PR introduces the concept of reserved cores more explicitly in the `node` code base.

Previously, one core was meant to be reserved for networking and general processing, while the remaining cores on the machine were used for data workers.  We now reserve more than 1 core depending on the number of cores available on the host, specifically:

- Hosts with 4 cores still reserve only one core.
- Hosts with 16 cores or less reserve 2 cores.
- Hosts with 32 cores or less reserve 3 cores.
- Hosts with 64 cores or less reserve 4 cores.
- Hosts with more than 64 cores reserve 5 cores.

The reason for this change is that during peak traffic moments, we require more than just one core to deal with the burst network traffic, validation, and OS requirements.

This change may be reverted by node operators by setting the `dataWorkerCount` setting in the `engine` section to the desired number of cores to be used by data workers. The setting will not produce effects if set above `GOMAXPROCS` (which by default is equal to the number of logical cores on the machine).

This PR also introduces the automatic setting of `GOGC` / `GOMEMLIMIT` if they are not explicitly provided for the master process.

### DIFF
--- a/node/execution/intrinsics/token/application/token_application.go
+++ b/node/execution/intrinsics/token/application/token_application.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"crypto"
 	"encoding/binary"
-	"runtime"
 	"sync"
 
 	"github.com/iden3/go-iden3-crypto/poseidon"
@@ -12,6 +11,7 @@ import (
 	"go.uber.org/zap"
 	"google.golang.org/protobuf/proto"
 	"source.quilibrium.com/quilibrium/monorepo/node/config"
+	qruntime "source.quilibrium.com/quilibrium/monorepo/node/internal/runtime"
 	"source.quilibrium.com/quilibrium/monorepo/node/p2p"
 	"source.quilibrium.com/quilibrium/monorepo/node/protobufs"
 	"source.quilibrium.com/quilibrium/monorepo/node/store"
@@ -366,7 +366,7 @@ func (a *TokenApplication) ApplyTransitions(
 	}
 
 	wg := sync.WaitGroup{}
-	throttle := make(chan struct{}, runtime.GOMAXPROCS(-1))
+	throttle := make(chan struct{}, qruntime.WorkerCount(0, false))
 	for i, transition := range set {
 		if transition == nil {
 			continue
--- a/node/main.go
+++ b/node/main.go
@@ -28,17 +28,6 @@ import (
 	"syscall"
 	"time"
 
-	"go.uber.org/zap"
-	"golang.org/x/crypto/sha3"
-	"google.golang.org/protobuf/proto"
-	"source.quilibrium.com/quilibrium/monorepo/node/execution/intrinsics/token"
-	"source.quilibrium.com/quilibrium/monorepo/node/execution/intrinsics/token/application"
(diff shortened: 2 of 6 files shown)


node/execution/intrinsics/token/application/token_application.go
```diff
@@ -366,7 +366,7 @@ func (a *TokenApplication) ApplyTransitions(
 	}
 
 	wg := sync.WaitGroup{}
-	throttle := make(chan struct{}, runtime.GOMAXPROCS(-1))
+	throttle := make(chan struct{}, qruntime.WorkerCount(0, false))
```
petricadaipegsp, line 369: I've further lowered the concurrency of the mint transition applications to leave some cores available for networking even while the frame is built.

node/main.go
```diff
@@ -384,24 +384,25 @@ func main() {
...
 	}
 
+	if nodeConfig.Engine.DataWorkerBaseListenMultiaddr == "" {
+		nodeConfig.Engine.DataWorkerBaseListenMultiaddr = "/ip4/127.0.0.1/tcp/%d"
+	}
+	if nodeConfig.Engine.DataWorkerBaseListenPort == 0 {
+		nodeConfig.Engine.DataWorkerBaseListenPort = 40000
+	}
+	if nodeConfig.Engine.DataWorkerMemoryLimit == 0 {
+		nodeConfig.Engine.DataWorkerMemoryLimit = 1792 * 1024 * 1024 // 1.75GiB
```
petricadaipegsp, line 394: Do we want to keep the 1.75 GiB default per worker memory limit ? In my experiments I've never seen it use _that much_ memory. Should we lower this to 1GiB perhaps ? Or even lower maybe.

CassOnMars: it will increase once workers are responsible for their own management of state

node/execution/intrinsics/token/application/token_application.go
CassOnMars: validation isn't running on light provers

node/main.go
scmart: While you are making this change, it would be nice to keep a ENV variable that could set the number of workers, as it would require less change for anyone using GOMAXPROCS to manage workers.

scmart:
While you are making this change, it would be nice to keep an ENV variable that could set the number of workers, as it would require less change for anyone using GOMAXPROCS to manage workers.

node/execution/intrinsics/token/application/token_application.go
petricadaipegsp: Yes, this change was targeted at the beacon itself. The reason is that I think we should leave some room for RPC processing in order to avoid the remote side drop RPCs while the beacon is applying transitions.

petricadaipegsp:
> While you are making this change, it would be nice to keep an ENV variable that could set the number of workers, as it would require less change for anyone using GOMAXPROCS to manage workers.

You can still use `GOMAXPROCS` if you so wish: just use `GOMAXPROCS=17` if you want 15 workers, and so forth. Although it is better to not fudge `GOMAXPROCS` and instead just use the dedicated data workers setting.

PR #372, opened 2024-11-20
Enable AutoNATv1 and NATPortMap

petricadaipegsp:
This PR enables two built-in libp2p functionalities:
- AutoNATv1, which allows nodes to check their external reachability via other peers.
- NATPortMap, which uses UPNP to automatically open node ports when available.

It also uses the built in libp2p reachability event emission (based on AutoNATv1) to print changes to the reachability. The log messages look as follows:

```
{"level":"info","ts":1732138895.894482,"logger":"reachability","caller":"p2p/blossomsub.go:333","msg":"node is externally reachable"}
```

### DIFF
--- a/node/p2p/blossomsub.go
+++ b/node/p2p/blossomsub.go
@@ -5,15 +5,10 @@ import (
 	"context"
 	"crypto/rand"
 	"encoding/hex"
-	"encoding/json"
 	"fmt"
-	"io"
 	"math/big"
 	"math/bits"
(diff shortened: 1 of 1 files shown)


petricadaipegsp:
The reachability status could be integrated in the future into the data peer announcements, such that peers first try to connect to public peers, and only then attempt private peers (in case they have direct connections). In general we should prioritize peers to which we already have connections.

PR #373, opened 2024-11-21
Use buffered channels when applicable

petricadaipegsp:
References https://github.com/QuilibriumNetwork/ceremonyclient/pull/357
References https://github.com/QuilibriumNetwork/ceremonyclient/commit/6e2c7f402854e77188a3a7c44fa75c6d21119b04

This PR brings back the buffered channels in the data time reel and data clock consensus engine, but only for channels where we are not supposed to drop the contents (i.e. every send to these channels has no `default` case).

The reason on why I've been advocating more for buffered channels is ordering - if we use an unbuffered channel, and each time when we want to write something to that channel we start a goroutine to write onto the channel, we have effectively an infinite set of buffers of size 1 (the goroutine itself, holding an item until the sending succeeds), but the order in which items will be written is undefined. The buffered channels bring back a bit of that ordering.

### DIFF
--- a/node/consensus/data/message_handler.go
+++ b/node/consensus/data/message_handler.go
@@ -14,14 +14,15 @@ import (
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/known/anypb"
 	"source.quilibrium.com/quilibrium/monorepo/node/config"
-	"source.quilibrium.com/quilibrium/monorepo/node/consensus"
 	"source.quilibrium.com/quilibrium/monorepo/node/execution/intrinsics/token/application"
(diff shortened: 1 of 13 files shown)


node/consensus/data/message_handler.go
```diff
@@ -47,26 +49,26 @@ func (e *DataClockConsensusEngine) runFrameMessageHandler() {
...
-				case protobufs.ClockFrameType:
-					if err := e.handleClockFrameData(
-						message.From,
-						msg.Address,
-						any,
-						false,
-					); err != nil {
-						return
-					}
+			switch any.TypeUrl {
```
petricadaipegsp, line 52: Frames, via PubSub, are rare enough that we can process them serially.

PR #385, opened 2024-11-24
Increase default maximum message size to 20MiB

petricadaipegsp:
This PR proposes increasing the PubSub default maximum message size to 20MiB preemptively, as frames are already close to the current limit of 10 MiB.

This change however is moderately risky, as un-updated nodes may drop these large messages. We will need a cutoff version (maybe dynamic - return 2.0.4 for 24 hours after the cutoff date, then 2.0.5).

### DIFF
--- a/go-libp2p-blossomsub/pubsub.go
+++ b/go-libp2p-blossomsub/pubsub.go
@@ -26,8 +26,8 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 )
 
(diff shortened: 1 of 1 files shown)


CassOnMars:
I went ahead and merged but I think the ideal strat here would be a "soft upper bound", e.g. 10MB as the upper limit for the merged RPCs, but for unmergeable messages due to size, hard limit of 20

PR #386, opened 2024-11-24
Always start sync from head

petricadaipegsp:
This PR simplifies the sync procedure to always start from the head of the data time reel, instead of a starting frame. The reason for this is that at the moment the data time reel head is always the 'furthest point in time'. This should avoid sync procedures that start in the past.

### DIFF
--- a/node/consensus/data/consensus_frames.go
+++ b/node/consensus/data/consensus_frames.go
@@ -21,12 +21,13 @@ import (
 
 const defaultSyncTimeout = 4 * time.Second
 
-func (e *DataClockConsensusEngine) collect(
-	enqueuedFrame *protobufs.ClockFrame,
-) (*protobufs.ClockFrame, error) {
+func (e *DataClockConsensusEngine) syncWithMesh() error {
 	e.logger.Info("collecting vdf proofs")
 
-	latest := enqueuedFrame
+	latest, err := e.dataTimeReel.Head()
+	if err != nil {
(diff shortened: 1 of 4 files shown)


node/consensus/data/consensus_frames.go
```diff
@@ -21,12 +21,13 @@ import (
 
 const defaultSyncTimeout = 4 * time.Second
 
-func (e *DataClockConsensusEngine) collect(
-	enqueuedFrame *protobufs.ClockFrame,
-) (*protobufs.ClockFrame, error) {
+func (e *DataClockConsensusEngine) syncWithMesh() error {
 	e.logger.Info("collecting vdf proofs")
 
-	latest := enqueuedFrame
```
CassOnMars, line 29: this has the possibility it may be behind and refetch already retrieved frames, we should use the max of enqueued frame or time reel head

adriansmares: That's exactly what the change does, right ?

petricadaipegsp: The enqueued frame, by the time it reaches collect, _was_ the head at some point. Collecting the head at this point is guaranteed to be at least equal or greater to the enqueued frame. But I'll refactor a bit.

CassOnMars: Ah right, makes sense – I was still thinking from the mental model where collect occurred at a prior callsite

PR #394, opened 2024-11-28
Change message size limits

petricadaipegsp:
This PR introduces the following changes:
- The built in sync RPC server message limits have been increased to 40 MiB, in order to allow larger frames to be sent via it.
- The soft max message size limit in the PubSub has been lowered to 1 MiB. Note that the soft limit only affects batched messages (like `IHAVE`, `IWANT`, `IDONTWANT` and multiple `PUBLISH` messages as a result of an `IWANT`).
  - This change improves latency by allowing messages to be received by the requesting peer faster, thus allowing this peer to emit `IDONTWANT` messages to its mesh, and to start heartbeating them as well.
  - Large messages are still allowed on the mesh, but longer term we need to get rid of them. They cause massive delays as they cannot be passed forward while being downloaded.

### DIFF
--- a/go-libp2p-blossomsub/pubsub.go
+++ b/go-libp2p-blossomsub/pubsub.go
@@ -26,8 +26,8 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 )
 
-// DefaultSoftMaxMessageSize is 10 MiB.
-const DefaultSoftMaxMessageSize = 10 << 20
+// DefaultSoftMaxMessageSize is 1 MiB.
+const DefaultSoftMaxMessageSize = 1 << 20
 
 // DefaultHardMaxMessageSize is 20 MB.
 const DefaultHardMaxMessageSize = 10 << 21
--- a/node/consensus/data/data_clock_consensus_engine.go
+++ b/node/consensus/data/data_clock_consensus_engine.go
@@ -331,8 +331,8 @@ func (e *DataClockConsensusEngine) Start() <-chan error {
 	e.pubSub.Subscribe(e.infoFilter, e.handleInfoMessage)
 	go func() {
 		server := qgrpc.NewServer(
-			grpc.MaxSendMsgSize(20*1024*1024),
-			grpc.MaxRecvMsgSize(20*1024*1024),
+			grpc.MaxSendMsgSize(40*1024*1024),
+			grpc.MaxRecvMsgSize(40*1024*1024),
 		)
 		protobufs.RegisterDataServiceServer(server, e)
 		if err := e.pubSub.StartDirectChannelListener(


petricadaipegsp:
Putting this on hold until we decide if this is the way we want to go forward.

go-libp2p-blossomsub/pubsub.go
```diff
@@ -26,8 +26,8 @@ import (
 	logging "github.com/ipfs/go-log/v2"
 )
 
-// DefaultSoftMaxMessageSize is 10 MiB.
-const DefaultSoftMaxMessageSize = 10 << 20
+// DefaultSoftMaxMessageSize is 1 MiB.
+const DefaultSoftMaxMessageSize = 1 << 20
 
 // DefaultHardMaxMessageSize is 20 MB.
 const DefaultHardMaxMessageSize = 10 << 21
```
CassOnMars, line 33: I'd argue we should increase the hard maximum to stay in sync with the grpc server, and have the grpc server use this value for consistency

petricadaipegsp: Increasing it is not panacea. The bigger we make it, the easier it is for potential attackers to create broadcast storms of bad messages with large sizes and to consume excessive bandwidth. All of the scoring in Gossip/BlossomSub are based on message counts / rates, and not bandwidth effectively spent on a message, under the assumption that large messages are not really what is broadcasted via PubSub (and instead the PubSub is used only to broadcast that the peer _has_ the data, not the data itself).

It also doesn't solve our large frame issue - middle peers which are not updated will still drop these jumbo frames, and it takes quite some time until peers are updated, and by that time, we can do fragmentation.

I personally want this value back to something like 2 or 4 MiB - it is way too big to be useful, and we see this today with the 17-18MiB messages that take a minute to propagate.

CassOnMars: wouldn't the problem persist with fragmentation, given we drop non-conformant messages?

petricadaipegsp: That's correct, and it is why I've used a different bitmask (4 leading zeroes) for the fragment bitmask.

There is no real backwards compatible option here:
- Frames bigger than 20 MiB can be synced only from the beacon, because the beacon has increased limits. `2.0.4-p2` has a 20 MiB limit (what this PR is about).
- Even if the PubSub limit is increased on the beacon, and the middle nodes are also updated, if the receiver is not updated, they cannot receive the frame whose size exceeds 20MiB.
- The only case in which an out of date node can download a frame, is if they download it from an updated node with increased limits (what this PR proposes).

Even nodes which are altruistic enough today to set their rate limits to more than 10 unique peers per minute cannot send frames above 20MiB, due to the gRPC message size limits.

PR #396, opened 2024-11-28
Support frame fragmentation and dissemination

petricadaipegsp:
This PR introduces a frame fragmentation system based on Reed-Solomon error correction codes.

Currently, frames have large sizes (~20MiB, sometimes even more), which limits how fast they propagate around the mesh. As a reminder, a message cannot be broadcasted to nearby peers until it is fully transferred and validated. While we could keep on increasing the max message size, it won't help much with the propagation speed.

Instead, what we could do is to split the binary representation into multiple shards, apply RS EC, and propagate the resulting shards individually. As they are smaller messages, they should propagate faster throughout the mesh, and losses are handled more gracefully.

The beacon is responsible for publishing the fragments. It can choose multiple styles of publication via configuration. Available options are:
  - `full` - publish only the full frame.
  - `fragmented` - publish only fragmented form of the full frame.
  - `dual` - publish both fragments and the full frame.
  - `threshold` - publish the full frame if the size of the frame is below a certain threshold, or fragment the frame otherwise.

The fragmentation parameters (the number of data shards and parity shards) are self contained - the sender (the beacon) can alter them on demand, and the receiver will determine them automatically from the fragment contents.


petricadaipegsp:
I've ran tests in the past 24 hours with these changes, by having a server re-transmit the frame in fragmented form on the main net. The RS data / parity shard, and the hash functions have been chosen based on this testing in order to minimize latency.

Synthetic benchmarks are available for these parameters in the `fragmentation` package. Generally, the latency is very sensitive to the hashing function, and `sha256` has been the fastest cryptographic hash available (I did not test `blake3`, but I can do it if there is interest in doing so). The the fragmentation and re-assembly are not bottlenecks.

PR #416, opened 2025-01-23
Rust module dependencies version bump

blacks1ne:
Bumps version of Rust module dependencies.
UniFFI version cannot be updated due to missing Go bindings.
There seems to be no impact on the performance.


CassOnMars:
will rebase this and include it in the same update with the upgrade to v0.3.0 of uniffi-bindgen-go

PR #422, opened 2025-01-24
Fix Docker warning about FROM and AS being in different letter case

blacks1ne:


winged-pegasus:
Resolved by #429.  I merged with my changes, thank you!

PR #424, opened 2025-03-29
improve docker build flow

winged-pegasus:


winged-pegasus:
This may need some rework, feel free to reopen if it was an overall improvement for you.

PR #425, opened 2025-03-30
improve compile error feedback

PalanQu:


CassOnMars:
This is great! Getting to a more formalized structure like this will help upstream consumers of the tooling. Thanks!

PR #453, opened 2025-10-26
Default to Info log level unless debug is enabled

blacks1ne:
Sets file logger's logging level to _Info_ unless **debug** parameter is enabled


CassOnMars:
Nice. I'm going to add an issue in reference to this, that we need to have flexible logging controls – right now, whenever theres an issue, we default to --debug which is _extremely_ noisy. it would be nicer if we could instead set log levels for different components

PR #455, opened 2025-10-28
issue: differences in double-width character rendering in specific terminal may mis-scale Q text logo

blacks1ne:


config/config.go
```diff
@@ -459,7 +459,7 @@ func Flags(blockchar **string, ver **string) {
 		setter{
 			*blockchar,
 			*ver,
-			"\U0001F1EB\U0001F1F7 ",
+			"\U0001F1EB\U0001F1F7",
```
CassOnMars, line 462: We have the space there because a lot of the terminal emulators don't correctly handle combining unicode characters correctly, causing it to treat the two unicode points as one character (correct), but has a double width (also correct), but does not _apply_ double width to the character when printing, resulting in overlapping flags. In some cases, terminals need the extra space, in others, they don't. If we're keeping this easter egg for longer than 2.1.0, we should probably find a way to determine how double-width characters render on a given terminal. Note: I don't know if there's a standard way to do that, many terminal emulators have issues that crop up around this from time to time, and are prioritized low or "not fixing" (e.g. https://github.com/microsoft/terminal/issues/16852)

PR #456, opened 2025-10-29
Missing "keys.yml" re-creation fix

blacks1ne:
If `key.keyManagerFile.createIfMissing` is set to `true` and the `keys.yml` file is missing, the node is expected to re-create it upon startup.
Currently, the node panics trying to decode YAML from empty file.
The fix skips the decoding if the `keys.yml` file size is zero (implying it is being re-created).


CassOnMars:
Awesome, thanks!

PR #457, opened 2025-10-29
v2.1.0.5

CassOnMars:


consensus/counters/strict_monotonic_counter.go
```diff
@@ -0,0 +1,50 @@
+package counters
+
+import "sync/atomic"
+
+// StrictMonotonicCounter is a helper struct which implements a strict monotonic
+// counter. StrictMonotonicCounter is implemented using atomic operations and
+// doesn't allow to set a value which is lower or equal to the already stored
+// ne. The counter is implemented solely with non-blocking atomic operations for
```
winged-pegasus: one?

winged-pegasus: StrictMonotonicCounter is implemented using atomic operations and
// doesn't allow to set a value which is lower or equal to the already stored
// ne. <-- value?

PR #458, opened 2025-10-29
Print worker filter on start and skip respawn for empty filter loaded from store

blacks1ne:
Worker Manager prints the following messages upon node start when trying to respawn worker after loading worker from store. 
```json
{"level":"error","ts":1761757093.1862228,"logger":"worker_manager","caller":"worker/manager.go:555","msg":"could not respawn worker","process":"master","error":"rpc error: code = Unavailable desc = connection error: desc = \"transport: Error while dialing: dial tcp 127.0.0.1:60000: connect: connection refused\"","stacktrace":"source.quilibrium.com/quilibrium/monorepo/node/worker.(*WorkerManager).loadWorkersFromStore\n\t/opt/ceremonyclient/node/worker/manager.go:555\nsource.quilibrium.com/quilibrium/monorepo/node/worker.(*WorkerManager).Start\n\t/opt/ceremonyclient/node/worker/manager.go:107\nsource.quilibrium.com/quilibrium/monorepo/node/consensus/global.(*GlobalConsensusEngine).Start\n\t/opt/ceremonyclient/node/consensus/global/global_consensus_engine.go:461\nsource.quilibrium.com/quilibrium/monorepo/node/app.(*MasterNode).Start\n\t/opt/ceremonyclient/node/app/node.go:87\nmain.main.func6\n\t/opt/ceremonyclient/node/main.go:561"}
```
The main motivation for the wotker respawn is setting non-empty filter that is used on worker start by default.
At the time of the check, the filters fetched from the store are empty, thus the worker respawn makes no sense.
The PR skip worker respawn if the worker loaded from store has empty filter and prints worker filter on its start to clarity.

### DIFF
--- a/node/datarpc/data_worker_ipc_server.go
+++ b/node/datarpc/data_worker_ipc_server.go
@@ -193,6 +193,7 @@ func (r *DataWorkerIPCServer) RespawnServer(filter []byte) error {
 		"data worker listening",
 		zap.String("address", r.listenAddrGRPC),
 		zap.String("resolved", lis.Addr().String()),
+		zap.ByteString("filter", filter),
 	)
 	if len(filter) != 0 {
 		globalTimeReel, err := r.appConsensusEngineFactory.CreateGlobalTimeReel()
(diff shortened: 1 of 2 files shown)


CassOnMars:
We want the respawn to happen with a nil filter, because it can be the case that a worker is evicted from a shard. In that case, we don't want the worker to continue emitting unnecessary messages or serving as a stale replica.

PR #498, opened 2025-12-19
add port checks to workers

winged-pegasus:


blacks1ne:
Could you please explain the challenge and the approach?

winged-pegasus:
The challenge is that the workers are trying to restart and they can't because the port is already in use.

The approach here is that the workers will First, check the ports to see if they're available and if they are then to proceed but otherwise back off and wait until the time out of 30 seconds and if it doesn't hit 30 seconds without them being available then they will just exit.

Upon further diagnosis, I'm not 100% sure this is the entire problem, but it does seem at least 90% effective.

The remaining may be due to the IPC server restarting on exit. I need to confirm this on a develop branch.

If this is the case, then a more simple solution is to add a second check, which is more of a band-aid, and then if the check fails, that the IPC server is still running, then to just stop it. But the better approach would be to figure out why it's restarting and fix that.

Again, I need to verify that this is the case if it wasn't just something that happens on my server after my these changes in this PR.

If this hat is the case then it would be interesting to know if this is the node worker or the master process initiating the IPC server restart on shutdown.

If you want to verify this, it would be adding logs to the IPC server and worker start/stop and adding a log for checking when an interrupt for the server happens.

The other thing you'll need to look for is the IPC server saying that it's listening. I know that on mine I saw that several of my cores would restart but not necessarily start listening before the shutdown was complete.

May be easier to diagnose this on or replicate this on a larger server with more cores, but I know that my 9950X was seeing this issue, and if this was the case, then it's not isolated to larger servers.

-------- Original Message --------

blacks1ne:
Thanks for the clarifications.
1. Frankly speaking, I have not spotted any issues with workers on my side. But I am running standalone nodes. Are you also running a standalone node or a cluster?
4. As you still seem to be experimenting, I would mark this PR as a **DRAFT**, implying it is not ready to be merged yet.
5. If this feature could benefit only some node, it could be worthy to make it configurable instead of forcing default behavior.
6. Once you complete the experiments and finalize the code, I could share some suggestions. For example, you have `// Log but don't fail - port was available`, but are not actually logging anything :)

PR #501, opened 2025-12-24
feat: add comprehensive docker guide and add config generator via cli

hhamud:
* adds a config gen cli
* moves all docker files into docker folder
* updates readme in docker


blacks1ne:
Thanks for your work on the Docker side and this PR. I guess we all agree that Docker instructions have long been outdated and it would be nice to update them while enhancing their quality.

[1] The only question generic question is what information should be kept in the repository instead of the official documentation available at https://docs.quilibrium.com. I trust @CassOnMars will provide guidance on that. Anyway, it is not a blocking question. 

As a disclaimed, I must admit that I am  a) a community member/volunteer and b) not a Docker expert. However, I hope that my comments may be useful before @CassOnMars as a project owner/lead (or someone she delegates to) has the time to review this PR.

Speaking of the repository structure changes, as I understand, you're suggesting to move all Docker-related files from the repository root to the `docker` directory. On one hand, this makes sense making the repository root cleaner. On the other hand, this introduces the path challenge as the Docker root no longer matches the repository root.
[2] It seems the Docker-related instructions and tasks are updated to go up one directory to reach the repository root. This assumes the task will be run from the `docker` directory and not from the repository root as `task --taskfile docker/Taskfile.yml`. However, I do not see any guideline in `docker/README.md` on the expected working directory when invoking tasks. It could be useful to add it.
[3] Another point of attention is that Docker-related tasks exclude MacOS node and qclient binary compilation which is done directly on MacOS bypassing Docker. That's why the original Taskfile is split into the repository root `Taskfile.yml` and the Docker-specific one in `docker/Taskfile.yml`. But if so, what's the value of having MacOS-specific `build_node_arm64_macos` and  `build_qclient_arm64_macos` task definitions in `docker/Taskfile.yml` while they are not Docker-related and are already present in the repository root `Taskfile.yml` ?

I'll comment on specific file changes in a separate comment.

blacks1ne:
Proceeding to Dockerfiles (not added in this PR, but still requiring a cleanup):
[4] I suppose `Dockerfile.conntest.source` was created for some tests and does not serve any valuable purpose, thus can be safely removed. @CassOnMars please confirm. If it is useful, maybe we could document its purpose and use cases.
[5] I believe `Dockerfile.vdf.*.source` files were used for VDF benchmarks during the summer of 2025 and are also safe to remove. @CassOnMars please confirm. If it is useful, maybe we could document its purpose and use cases.
[6] `Dockerfile.release` seems to aim to build a Docker image from the official/signed binaries. However, it a) expects to have node binaries pre-downloaded to the `node` directory and b) compiles qclient instead of using official binaries. The official signed `node` and `qclient` binaries with digests and signatures should be downloadable via https://releases.quilibrium.com/release and https://releases.quilibrium.com/qclient-release respectively.
[a] I believe the Docker instructions in this Dockerfile should be updated to fetch the official release files.
[b] I am not sure why we should use `grpcurl` version `1.9.1` instead of `latest`.
[c] I wonder what is the benefit of relying on Alpine Linux while the other Docker files and official instructions recommend Ubuntu.

blacks1ne:
Concluding with `docker/README.md` :

As far as I understad, we use Docker for the following purposes (please validate):
1) compile/build Linux AMD64 or ARM64 `node` and/or `qclient` binaries
2) create Docker images with the official release binaries
3) create Docker images with the custom `node` and `qclient` binaries
4) run Docker images with official or custom `node` and `qclient` binaries
5) run `node` and `qclient` binaries via Docker

[7] Section `1. System Preparation` is not required for purposes 1)-3), overlaps with documentation at https://docs.quilibrium.com and contains questionable statements:
[a] It claims Quilibrium relies on `QUIC (UDP)`, while QUIC is a protocol build over UDP, thus `QUIC/UDP` like `TCP/IP` not being `TCP (IP)`. Besides, P2P ports support TCP while stream/RPC ports support only TCP.
[b] Firewall configuration is wrong:
- ports 22, 443, 8337, 8338 should not be exposed by default
- port 8340 should be TCP
- port 8336 is restricted to UDP while it can also be TCP
- ports 50000+ are limited to 10 implying only 11 workers maximum which is not sufficient even for middle-class laptops
- ports 50000+ are mentioned twice: once with UDP and once with TCP, while only one of the two will be needed, and if both are desired it is simple to omit protocol specification
- ports 60000+/TCP are not included (workers will not work properly)

[8] (Enhancement) Cross-compilation instructions are missing are missing how to enable cross-compilation on the local OS and in Docker.

[9] (Enhancement) `Customizing Configuration` does not allow to customize the following:
- master p2p protocol
- master stream port
- worker p2p ports and protocol
- worker stream ports

hhamud:
> Proceeding to Dockerfiles (not added in this PR, but still requiring a cleanup): [4] I suppose `Dockerfile.conntest.source` was created for some tests and does not serve any valuable purpose, thus can be safely removed. @CassOnMars please confirm. If it is useful, maybe we could document its purpose and use cases. [5] I believe `Dockerfile.vdf.*.source` files were used for VDF benchmarks during the summer of 2025 and are also safe to remove. @CassOnMars please confirm. If it is useful, maybe we could document its purpose and use cases. [6] `Dockerfile.release` seems to aim to build a Docker image from the official/signed binaries. However, it a) expects to have node binaries pre-downloaded to the `node` directory and b) compiles qclient instead of using official binaries. The official signed `node` and `qclient` binaries with digests and signatures should be downloadable via https://releases.quilibrium.com/release and https://releases.quilibrium.com/qclient-release respectively. [a] I believe the Docker instructions in this Dockerfile should be updated to fetch the official release files. [b] I am not sure why we should use `grpcurl` version `1.9.1` instead of `latest`. [c] I wonder what is the benefit of relying on Alpine Linux while the other Docker files and official instructions recommend Ubuntu.

Couldn't we just get them to build from source rather than downloading the release binaries?

hhamud:
> Concluding with `docker/README.md` :
> 
> As far as I understad, we use Docker for the following purposes (please validate):
> 
>     1. compile/build Linux AMD64 or ARM64 `node` and/or `qclient` binaries
> 
>     2. create Docker images with the official release binaries
> 
>     3. create Docker images with the custom `node` and `qclient` binaries
> 
>     4. run Docker images with official or custom `node` and `qclient` binaries
> 
>     5. run `node` and `qclient` binaries via Docker
> 
> 
> [7] Section `1. System Preparation` is not required for purposes 1)-3), overlaps with documentation at https://docs.quilibrium.com and contains questionable statements: [a] It claims Quilibrium relies on `QUIC (UDP)`, while QUIC is a protocol build over UDP, thus `QUIC/UDP` like `TCP/IP` not being `TCP (IP)`. Besides, P2P ports support TCP while stream/RPC ports support only TCP. [b] Firewall configuration is wrong:
> 
>     * ports 22, 443, 8337, 8338 should not be exposed by default
> 
>     * port 8340 should be TCP
> 
>     * port 8336 is restricted to UDP while it can also be TCP
> 
>     * ports 50000+ are limited to 10 implying only 11 workers maximum which is not sufficient even for middle-class laptops
> 
>     * ports 50000+ are mentioned twice: once with UDP and once with TCP, while only one of the two will be needed, and if both are desired it is simple to omit protocol specification
> 
>     * ports 60000+/TCP are not included (workers will not work properly)
> 
> 
> [8] (Enhancement) Cross-compilation instructions are missing are missing how to enable cross-compilation on the local OS and in Docker.
> 
> [9] (Enhancement) `Customizing Configuration` does not allow to customize the following:
> 
>     * master p2p protocol
> 
>     * master stream port
> 
>     * worker p2p ports and protocol
> 
>     * worker stream ports

tbh I had constant issues with connections until I had opened up all those ports.

winged-pegasus:
There is value for node runners to be able to run VDF tests independently.  Most, I think are not aware of the utility/benefit, but it is a good way to get a benchmark on servers.  

How much of a value?  I don't know, considering that you need access to these servers to even run the test, but if you were to get a rental or test/trial period it would be an easy way to run benchmarks across servers to validate how well a server is at VDF tests even across your own laptops/desktops/servers (mobile devices?).  I would not delete this Dockerfile just to clean up.

The conntest Dockerfile, I am not really sure what use it has for a normal user beyond a normal node being able to run. I'm assuming it is a hyper-specialized tool that Cassie was using for testing 2.1 builds. 

As for the Darwin task(s), this should be able to be used while on MacOS, so for most people it is not useful as they are building for Linux, I would not go so far as to remove it. Maybe prevent it from being run on non-Darwin systems, but I wouldn't out-right delete it. If it is not even useful for building on MacOS, I would probably just leave the task there, but script the build for Darwin systems so the user can still use task to start the build process regardless of system architecture/OS-- while updating the build documentation to note why and which uses Docker vs something else (mostly a meta/developer note for the future rather than a builder concern). 

As for cleaning up Docker files into their own directory, I can get behind this, but I would leave the Taskfile.yml in the root directory and update the file references to point to the updated Dockerfile locations.  This may require updating the docker file COPY commands to go up one directory, but not a deal breaker as it's a one time thing and it doesn't change the flow of:

1. `sudo apt install task`
2. `task <command>`

for someone to build.  

If somebody is aiming to run a node inside a Dockerfile, we should actually deal with making a docker image that just pulls the distributed binaries and tag it with a version.  No need to build, unless they absolutely want to, which is why the source Dockerfile is useful-- as some do want to build from source. 

As for REQUIRED PORTS:
1. If a user wants to run a node inside Docker, there is the challenge that we're dealing with a container;  the ports the node is using must be either mapped from internal Docker ports to the host's ports.  This is easily done with a docker-compose file or attempting to open/expose ports in the Dockerfile if the amount of workers is known but is not really that flexible enough to deal with dynamically making it "just work".  This would imply we need some script to expose ports dynamically based on the number of CPU threads the container is using/has access to or use an override, e.g. env variable to pass into the docker file (WORKERS=10).  The docker file must be aware of the base ports for p2pListen/streaming as well, which can change based on the config file.  And this doesn't take into account if the user has manually defined their workers (another can of worms). 
2. You need 8336/8340 open at a minimum, by default, however this assumes no workers.  Many people will also change their port to be non-standard. 
3. if there are workers, we can't limit the number of workers-- it must be dynamic enough to cover all the workers, whether it be 3 or 251.  There is a non-negligible amount of runners who use more than 10 workers. 

Other thoughts:
1. we can probably use grpcurl:latest 
2. I do not think that the distro really matters all that much for Linux operations-- as long as it is pulling the right binary.  There is a difference in OS size: Ubuntu is several gigs, whereas Alpine is much smaller (~130MB).  The main point here is that if it works, it works.  If a user has a specific requirement beyond the provided tooling, they are free to modify it based on their requirements. If they don't know the difference, then as long as it functions, then it won't matter.  The biggest con is mostly that if they need help diagnosing or get help, most assume they are running on Ubuntu.  

This is where a standardized, comprehensive tooling set is useful that will set up and contain diagnostic tools that run for a user regardless of Linux OS (or even Darwin, if we get ambitious).  This is well beyond the scope of this PR, so I won't dive into it here, but this is the route we must go if we aim to make it truly world-class and friction-less as possible for the most people. (There will always be "I'm unique" scenarios, but they usually can figure it out themselves or be less common).

winged-pegasus:
If you're interested, I have a massive collection of scripts here to illustrate how complex it can get: https://github.com/tjsturos/qtools/tree/main

hhamud:
> If you're interested, I have a massive collection of scripts here to illustrate how complex it can get: https://github.com/tjsturos/qtools/tree/main

really nice collection of scripts, bash gets wieldy as it grows and converting them into a rust tui program would be the way to go.

winged-pegasus:
That's what I realized after the fact. I was initially aiming at making commands native for Linux CLI as a person would type in on the CLI, but over time I found it to be cumbersome as it got much bigger.

I agree, I was thinking it would be worth taking my repo and using AI to convert to a TUI and using/serving as an official toolset. (For context, I started this set pre-hiring). It is a great base for reference, and there are a few learning lessons there, as well.

Some of them are: there are so many commands, auto-complete is necessary, although in a TUI, it would mean a categorization of commands so you can filter by need and using branching hotkeys. Documentation is necessary, so the user who is not familiar with commands understands what they do and when to use them. Access next to the command would be useful (not just in a separate docs page) so users can get immediate help rather than refer to a web page. Parameters need to be documented as well, however with a TUI, it's reduces the need as you can prompt the user better as they go.

Another thing is, that this should be CLI compatible, not just a TUI, because not everyone is going to want to treat their nodes like a pet and the package may be used as an automated install, not just user interactive.

I initially was thinking to convert to Go, but going into Rust may be fine as well. Not really sure there is any reason to choose one or the other, the main point to a conversion would be to make a binary to distribute/install.

hhamud:
> There is value for node runners to be able to run VDF tests independently. Most, I think are not aware of the utility/benefit, but it is a good way to get a benchmark on servers.
> 
> How much of a value? I don't know, considering that you need access to these servers to even run the test, but if you were to get a rental or test/trial period it would be an easy way to run benchmarks across servers to validate how well a server is at VDF tests even across your own laptops/desktops/servers (mobile devices?). I would not delete this Dockerfile just to clean up.
> 
> The conntest Dockerfile, I am not really sure what use it has for a normal user beyond a normal node being able to run. I'm assuming it is a hyper-specialized tool that Cassie was using for testing 2.1 builds.
> 
> As for the Darwin task(s), this should be able to be used while on MacOS, so for most people it is not useful as they are building for Linux, I would not go so far as to remove it. Maybe prevent it from being run on non-Darwin systems, but I wouldn't out-right delete it. If it is not even useful for building on MacOS, I would probably just leave the task there, but script the build for Darwin systems so the user can still use task to start the build process regardless of system architecture/OS-- while updating the build documentation to note why and which uses Docker vs something else (mostly a meta/developer note for the future rather than a builder concern).
> 
> As for cleaning up Docker files into their own directory, I can get behind this, but I would leave the Taskfile.yml in the root directory and update the file references to point to the updated Dockerfile locations. This may require updating the docker file COPY commands to go up one directory, but not a deal breaker as it's a one time thing and it doesn't change the flow of:
> 
>     1. `sudo apt install task`
> 
>     2. `task <command>`
> 
> 
> for someone to build.
> 
> If somebody is aiming to run a node inside a Dockerfile, we should actually deal with making a docker image that just pulls the distributed binaries and tag it with a version. No need to build, unless they absolutely want to, which is why the source Dockerfile is useful-- as some do want to build from source.
> 
> As for REQUIRED PORTS:
> 
>     1. If a user wants to run a node inside Docker, there is the challenge that we're dealing with a container;  the ports the node is using must be either mapped from internal Docker ports to the host's ports.  This is easily done with a docker-compose file or attempting to open/expose ports in the Dockerfile if the amount of workers is known but is not really that flexible enough to deal with dynamically making it "just work".  This would imply we need some script to expose ports dynamically based on the number of CPU threads the container is using/has access to or use an override, e.g. env variable to pass into the docker file (WORKERS=10).  The docker file must be aware of the base ports for p2pListen/streaming as well, which can change based on the config file.  And this doesn't take into account if the user has manually defined their workers (another can of worms).
> 
>     2. You need 8336/8340 open at a minimum, by default, however this assumes no workers.  Many people will also change their port to be non-standard.
> 
>     3. if there are workers, we can't limit the number of workers-- it must be dynamic enough to cover all the workers, whether it be 3 or 251.  There is a non-negligible amount of runners who use more than 10 workers.
> 
> 
> Other thoughts:
> 
>     1. we can probably use grpcurl:latest
> 
>     2. I do not think that the distro really matters all that much for Linux operations-- as long as it is pulling the right binary.  There is a difference in OS size: Ubuntu is several gigs, whereas Alpine is much smaller (~130MB).  The main point here is that if it works, it works.  If a user has a specific requirement beyond the provided tooling, they are free to modify it based on their requirements. If they don't know the difference, then as long as it functions, then it won't matter.  The biggest con is mostly that if they need help diagnosing or get help, most assume they are running on Ubuntu.
> 
> 
> This is where a standardized, comprehensive tooling set is useful that will set up and contain diagnostic tools that run for a user regardless of Linux OS (or even Darwin, if we get ambitious). This is well beyond the scope of this PR, so I won't dive into it here, but this is the route we must go if we aim to make it truly world-class and friction-less as possible for the most people. (There will always be "I'm unique" scenarios, but they usually can figure it out themselves or be less common).

1. I'll get rid of the docker/taskfile and just use the main taskfile in the project repo. 
2. Not even sure if the docker compose file is even needed. 
3. for the ports networking, the docs has something that can use https://docs.docker.com/engine/network/drivers/host/ 
4. Dockerfiles can do with some more improvement definitely.

winged-pegasus:
[Platform support](https://docs.docker.com/engine/network/drivers/host/#platform-support)

The host networking driver is supported on:

- Docker Engine on Linux
- Docker Desktop version 4.34 and later (requires enabling the feature in Settings)
This looks like it wouldn't be an option for MacOS

winged-pegasus:
The docker compose file probably isn't needed, as you can directly pass args in the CLI command, I believe, and if the build/test commands are always predefined in the task file, we can just avoid the abstraction.

blacks1ne:
> Couldn't we just get them to build from source rather than downloading the release binaries?

I believe the benefit of the official/signed binaries towards less technical people is that these official binaries have the following attributes:
1. work as expected (assuming extensive testing done in CI/CD pipeline which is not necessarily the case yet)
2. have a publicly available source code version they were compiled from.

Speaking of 2, the implicit expectations is that if one compiles from source, he/she does the 1st level of troubleshooting before reporting issues. However, if one runs an official build, issues can be reported immediately.

So I still see value in using official/signed binaries.

blacks1ne:
> tbh I had constant issues with connections until I had opened up all those ports.

My point is that you accomplished the following
1. opened sufficient ports for the master to operate properly
2. opened more ports that necessary exposing your host to security threats
3. did not open enough ports for the workers to operate normally, something you will learn only once your node tries to enroll workers in shards

blacks1ne:
> There is value for node runners to be able to run VDF tests independently. Most, I think are not aware of the utility/benefit, but it is a good way to get a benchmark on servers.
> 
> How much of a value? I don't know, considering that you need access to these servers to even run the test, but if you were to get a rental or test/trial period it would be an easy way to run benchmarks across servers to validate how well a server is at VDF tests even across your own laptops/desktops/servers (mobile devices?). I would not delete this Dockerfile just to clean up.

VDF performance was definitely the key metric for node versions 1.4-2.0. However, I recall @CassOnMars saying that its impact in rewards will get significantly pushed back with 2.1. As you recall, @CassOnMars did ask us to run VDF benchmarks in summer'25. Yet, we still have no clear interpretations that could drive any decision making. We're even not sure if AVX512 builds provide any advantage.
If we were to keep the VDF benchmark Dockerfiles, maybe we could consolidate them into a single Dockerfile to clean up, if possible.

blacks1ne:
> > As for cleaning up Docker files into their own directory, I can get behind this, but I would leave the Taskfile.yml in the root directory

>     1. I'll get rid of the docker/taskfile and just use the main taskfile in the project repo.

Makes perfect sense

blacks1ne:
>     1. we can probably use grpcurl:latest

I would argue, `node` and `qclient` binaries provide sufficient actions to a general user. `grpcurl` is an advanced tool that very few users can benefit from for advanced debugging. If so, it makes no sense to keep it in a generic Docker file.
But if we keep it, `grpcurl@latest` is definitely the way to go.

blacks1ne:
> The biggest con is mostly that if they need help diagnosing or get help, most assume they are running on Ubuntu.

Supportability is my main concern

blacks1ne:
> If somebody is aiming to run a node inside a Dockerfile, we should actually deal with making a docker image that just pulls the distributed binaries and tag it with a version. No need to build, unless they absolutely want to, which is why the source Dockerfile is useful-- as some do want to build from source.

I believe @CassOnMars planned on making an official Docker image available alongside the official/signer binaries, but it would be great to confirm this

blacks1ne:
I wonder if we should not split this discussion into several issues, one per topic, as it is getting complex to follow

winged-pegasus:
> > There is value for node runners to be able to run VDF tests independently. Most, I think are not aware of the utility/benefit, but it is a good way to get a benchmark on servers.
> > How much of a value? I don't know, considering that you need access to these servers to even run the test, but if you were to get a rental or test/trial period it would be an easy way to run benchmarks across servers to validate how well a server is at VDF tests even across your own laptops/desktops/servers (mobile devices?). I would not delete this Dockerfile just to clean up.
> 
> VDF performance was definitely the key metric for node versions 1.4-2.0. However, I recall @CassOnMars saying that its impact in rewards will get significantly pushed back with 2.1. As you recall, @CassOnMars did ask us to run VDF benchmarks in summer'25. Yet, we still have no clear interpretations that could drive any decision making. We're even not sure if AVX512 builds provide any advantage. If we were to keep the VDF benchmark Dockerfiles, maybe we could consolidate them into a single Dockerfile to clean up, if possible.

VDF performance is not a straight link to rewards, mostly due to the VDF for proofs will adjust the difficulty based on the shard average and device capability in 2.1.  That said, if you only look at how generations work, when the global VDF difficulty hits a threshold, it will trigger the new generation.  With this in mind, it will be useful to find out what your node can actually do if it seems to be getting kicked out of shards or when trying to see where your hardware matches against current standards.

Yes, right now, not so useful, but in the long run, it will be useful for testing/diagnosing hardware, avx512 or not.

Side note: IIRC the AVX512 performances did not net a huge improvement in most systems, however some did benefit more than others... IMO this should be retested to see if this still is the case considering how much changes have been done.

winged-pegasus:
> I wonder if we should not split this discussion into several issues, one per topic, as it is getting complex to follow

could probably move some of these to their own issues

winged-pegasus:
> > The biggest con is mostly that if they need help diagnosing or get help, most assume they are running on Ubuntu.
> 
> Supportability is my main concern

yes, it makes sense to base it all off Ubuntu or Alpine, but considering most are going to be installing Ubuntu and most providers have Ubuntu ISOs for installation as options, then we should just standardize off Ubuntu 24.

winged-pegasus:
> If we were to keep the VDF benchmark Dockerfiles, maybe we could consolidate them into a single Dockerfile to clean up, if possible.

Sure, however, git history makes restoring them relatively easy if we need to resurrect them, however any easy win is simply moving them into a `docker/vdf/` directory.

hhamud:
and thy founder hath spoken

image

blacks1ne:
> > If we were to keep the VDF benchmark Dockerfiles, maybe we could consolidate them into a single Dockerfile to clean up, if possible.
> 
> Sure, however, git history makes restoring them relatively easy if we need to resurrect them, however any easy win is simply moving them into a `docker/vdf/` directory.

or to a distinct branch, e.g. `vdf-perf-test`, though separating in a branch is more likely to facilitate the code falling behind mainstream

Taskfile.yaml
```diff
@@ -55,67 +55,108 @@ tasks:
...
+
+  test:port:
+    desc: Test if the P2P port is visible to the world.
+    preconditions:
+      - sh: 'test -x "$(command -v nc)"'
+        msg: 'nc is not installed, install with "sudo apt install netcat"'
+      - sh: 'test -n "$NODE_PUBLIC_NAME"'
+        msg: 'The public DNS name or IP address of the server must be set in NODE_PUBLIC_NAME.'
+    cmds:
+      - 'nc -vzu ${NODE_PUBLIC_NAME} ${QUILIBRIUM_P2P_PORT:=8336}'
```
CassOnMars, line 117: nc doesn't work particularly well for this - we have a libp2p ping test utility here: https://github.com/QuilibriumNetwork/monorepo/blob/develop/go-libp2p/examples/p2p-ping/main.go

This is essentially the same logic being used for the actual reachability test

CassOnMars:
LGTM - the connection test will likely not be super helpful, but the rest of it is great, will merge

PR #537, opened 2026-05-25
Add Rust CI with GitHub Actions

dazthecorgi:
Adds CI for Rust code with GitHub Actions to make it easier to contribute. There are three jobs:

- `check`: make sure code compiles without generating artifacts (`channelwasm` crate is excluded, bc it doesn't compile currently).
- `test`: runs tests with [nextest](https://nexte.st/) (some tests are failing currently, these are temporarily excluded -- see `.nextest.toml` for the list).
- `release`: runs `node/build.sh` on the three supproted supported platforms (`macos-arm64`, `linux-arm64`, `linux-x86`). Depends on `check` to save the trees. It's important to run this in addition to `check`, bc some build issues are platform specific and only show up in release mode due native dependencies.

In the future we'd want to add lints with Clippy and Rustfmt checks as well. I'll create issues for the follow up tasks.


crates/libp2p-identity/Cargo.toml
```diff
@@ -180,6 +180,9 @@ version = "0.5"
 [dev-dependencies.hex-literal]
 version = "0.4.1"
 
+[dev-dependencies.quickcheck]
+version = "1"
+
```
dazthecorgi, line 185: Had to add this to get the crate to compile in test mode

dazthecorgi:
There is a passing run in my fork on the latest commit: https://github.com/dazthecorgi/monorepo/pull/1/checks

The workflow will only run in this repo once this PR is merged.

dazthecorgi:
ci job (in progress) after rebasing on top of the latest .23 branch: https://github.com/dazthecorgi/monorepo/actions/runs/26408180635

PR #553, opened 2026-05-26
add verify_prover_join_vdf_accepts_real_proof test

### LINKED ISSUE #548: ProverJoin VDF verifier uses SHA-256 instead of SHA3-256
### Summary

`verify_prover_join_vdf` in the Rust execution layer computes the VDF
challenge with `sha2::Sha256`, but the signer (both Rust pipeline and Go
reference) uses `Sha3_256`. The challenge is the input to
`frame_prover.verify_multi_proof(...)`, so every legitimate ProverJoin proof
fails verification.

### Where

Archive node, validating an incoming ProverJoin message at frame
materialization time.

### Symptom

Archive logs:

```
quil_engine/src/frame_materializer.rs: skipping message that failed signature validation
  error: "invalid argument: global: signature verification failed"
  frame: 5
  request_type: "0x00000312"   (TYPE_MESSAGE_BUNDLE wrapping a ProverJoin)
```

The error wording is misleading — the engine wraps any `Ok(false)` from
`intrinsic.validate(...)` as "signature verification failed", whether the
failing step is the BLS signature, the VDF multi-proof, or something else
returning `false`. In this case it's the VDF.

### Control flow

1. Signer — Rust pipeline computes the VDF proof against a SHA3-256
   challenge:
   [`crates/quil-engine/src/prover_pipeline.rs`](https://github.com/QuilibriumNetwork/monorepo/blob/v2.1.0.23/crates/quil-engine/src/prover_pipeline.rs)
   (search for `Sha3_256::digest(&output)`).
2. Go reference — same:
   [`node/execution/intrinsics/global/global_prover_join.go`](https://github.com/QuilibriumNetwork/monorepo/blob/v2.1.0.23/node/execution/intrinsics/global/global_prover_join.go)
   (search for `sha3.Sum256(frame.Header.Output)`).
3. Rust verifier — uses SHA-256 instead:
   [`crates/quil-execution/src/global_intrinsic/verify.rs`](https://github.com/QuilibriumNetwork/monorepo/blob/v2.1.0.23/crates/quil-execution/src/global_intrinsic/verify.rs)
   — in `verify_prover_join_vdf`, search for
   `sha2::Sha256::digest(frame_output)`.
4. The mismatched challenge gets passed to
   `frame_prover.verify_multi_proof(&challenge, difficulty, &id_refs, &sol_refs)`,
   which deterministically returns `Ok(false)` for any legitimate proof.

Effect is universal: every non-archive node's ProverJoin is rejected at every
archive that runs the Rust verifier path, regardless of who signs.

### How we confirmed it wasn't the BLS sig

Added a `tracing::warn!` before each `return Ok(false)` in
`verify_prover_join_signatures` (the BLS check). The warns never fired — the
BLS function returned `Ok(true)` — so the `Ok(false)` had to be coming from a
later step, which narrowed it to `verify_prover_join_vdf`.

A BLS self-verify also added in `prover_pipeline.rs` (parallel to the
existing VDF self-verify) round-tripped cleanly on the signer, confirming the
bytes and keypair were fine.

### Suggested fix

In `verify_prover_join_vdf`:

```rust
use sha3::Digest;
let challenge: [u8; 32] = sha3::Sha3_256::digest(frame_output).into();
```

Add `sha3 = "0.10"` to `crates/quil-execution/Cargo.toml`.

Also add a regression test for this.

dazthecorgi:
Fixes #548

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4152,6 +4152,7 @@ dependencies = [
  "serde",
  "serde_json",
  "sha2 0.10.8",
+ "sha3 0.10.8",
  "tempfile",
  "thiserror 2.0.12",
  "tokio",
--- a/crates/quil-execution/Cargo.toml
+++ b/crates/quil-execution/Cargo.toml
@@ -30,6 +30,7 @@ num-traits = "0.2"
 hex = "0.4"
 prost = "0.13"
 sha2 = "0.10"
+sha3 = "0.10"
 bs58 = "0.5"
 metrics = "0.24"
 serde = { version = "1", features = ["derive"] }
@@ -38,6 +39,7 @@ serde_json = "1"
 [dev-dependencies]
 tempfile = "3"
 quil-hypergraph = { path = "../quil-hypergraph", features = ["test-utils"] }
+quil-crypto = { path = "../quil-crypto", features = ["vdf-prover"] }
 quil-engine = { path = "../quil-engine" }
 quil-p2p = { path = "../quil-p2p" }
 ed448-rust = { path = "../ed448-rust" }
--- a/crates/quil-execution/src/global_intrinsic/verify.rs
+++ b/crates/quil-execution/src/global_intrinsic/verify.rs
@@ -323,9 +323,11 @@ pub fn verify_prover_join_vdf(
     // 1. Structural validation
     let validation = validate_prover_join_structural(op, current_frame_number)?;
 
-    // 2. Compute challenge from frame output
-    use sha2::Digest;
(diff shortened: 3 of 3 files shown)


dazthecorgi:
https://github.com/QuilibriumNetwork/monorepo/pull/529/commits/27cd8c5df638ee414185173b43ca42d630694c44 has added the fix since I opened this PR, but I think it's still worth merging, because of the new test case.

dazthecorgi:
test failure is from the target branch

PR #567, opened 2026-06-14
fix transaction safety for hypergraph store writes

dazthecorgi:
A vertex-data persistence test failure surfaced a broader problem where hypergraph writes could escape their surrounding transaction: persisting for uncommitted frames, leaking via a read path, or leaving the in-memory tree out of sync with the store on retry.

- **fix vertex data roundtrip test and add new one aa5274b8cd3d1c95824b3e328f2b2f213715290e** 
  - fixes https://github.com/QuilibriumNetwork/monorepo/issues/539, this made me look deeper into hypergraph persistence.
- **use tx when saving vertex data db439927d0b818a354fa9cb3dca3b5f99094e153**  
  - thread the txn through `save_vertex_underlying` so that vertex data isn't persisted if the tx reverts.
- **make `LazyVectorCommitmentTree::commit ` retry-safe c6c6398a7af52fa51f9e49461dfd1c5af05114af:** 
  - defer clearing dirty state until `txn.commit()` succeeds, so an aborted txn leaves the tree dirty for a safe retry (covers all staged writes, not just vertex data).
- **make `compute_shard_root` read-only 61db77433762b8b1ec2e61071457d17b0a7c3a53:** 
  - extract `compute_root()` so root comparison on the hot master-stream path stops leaking writes via the `NoopTxn` fallback.
- **require `RocksTxn` for store writes 4b61ab595e8066255e64346945e71d6956186190:** 
  - remove the silent direct-write fallback that masked the above bug; all writers now use the concrete `RocksTxn` batch and error loudly on an unrecognized txn.

See individual commit messages for full detail.

### DIFF
--- a/crates/quil-rpc/src/hypergraph_sync_probe.rs
+++ b/crates/quil-rpc/src/hypergraph_sync_probe.rs
@@ -23,7 +23,7 @@ use quil_types::proto::application::{
     hypergraph_sync_query, hypergraph_sync_response, HypergraphPhaseSet,
     HypergraphSyncGetBranchRequest, HypergraphSyncGetLeavesRequest, HypergraphSyncQuery,
 };
-use quil_types::store::ShardKey;
+use quil_types::store::{HypergraphStore, ShardKey, Transaction};
 
 use crate::archive_client::{build_quil_client_config, QuilTlsConnector};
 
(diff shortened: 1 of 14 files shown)


crates/quil-rpc/src/hypergraph_sync_probe.rs
```diff
@@ -1039,22 +1039,52 @@ pub async fn ensure_shard_tree_fresh(
...
-        if hg_store
-            .save_vertex_underlying(set_str, phase_str, shard, &entry.key, &entry.underlying_data)
-            .is_ok()
-        {
-            persisted_vertices += 1;
-        }
+/// Persist a shard tree blob and its per-vertex underlying blobs in one
+/// transaction. Either everything commits or (on error) nothing does —
+/// the txn is dropped, which aborts the batch. Returns the number of
+/// per-vertex blobs staged.
```
dazthecorgi, line 1065: note that this is a change in behavior. previously partial failures were tolerated

PR #578, opened 2026-07-02
add devnet scenario runner

dazthecorgi:
## Overview

Adds a Docker-based development network harness that runs multiple Quilibrium nodes locally and helps test consensus under controlled network partitions. 

It currently spins up 4 archive nodes + 1 client node via `docker compose`, fronted by a proxy that intercepts gossip and gRPC traffic, applies network partitions at specified consensus ranks, and reports the run's outcome (frame liveness, safety, client enrollment) back to the orchestrator. If the test fails, logs from each node are saved to disk.

## Quick start

Make sure you have the Rust toolchain and a local Docker installation with compose, then run the following command from the repo root:

```bash
# single run with one partition at rank 1, stopping at frame 5
cargo run -p devnet -- single --verbose --stopframe=5 --dir=crates/devnet \
  --rank-partitions='[{"rank":1 ,"partition1":["archive-1","archive-2","archive-3"],"partition2":["archive-4"]}]'
```

To see how it can find consensus bugs, comment out the following line and rerun the test:

https://github.com/QuilibriumNetwork/monorepo/blob/88919e61ad56ca5e01b0f4027401a526edab2a3e/crates/quil-engine/src/consensus_glue.rs#L242

The test will now fail, because `archive-4` will never recover from being partitioned out at rank 1.

## More info

See the `devnet` crate's readme for more info.


dazthecorgi:
the ci failures are from the base branch and are fixed in https://github.com/QuilibriumNetwork/monorepo/pull/577

PR #579, opened 2026-07-02
Make BLS QC/TC verifier committee-mandatory

dazthecorgi:
## Summary

Make BLS QC/TC verifier committee-mandatory  to close timeout-path fail-open.

## Context

A peer's `TimeoutState` embeds a full QC (its newest QC) and an optional prior-rank TC — complete certificates the sender chose, reconstructed from the sender's bytes on receipt. Each carries an `AggregateSignature { public_key, signature, bitmask }`, and the aggregate `public_key` travels inside the cert. A raw BLS check only proves `signature` is valid over the canonical message under that transmitted `public_key` — a key the sender controls — so it is trivially satisfiable: a peer generates its own keypair, self-signs the canonical message, and sets `public_key` to its own key.

What makes a QC actually mean "a supermajority of the real committee signed" is `bind_aggregate_pubkey_to_committee`: it reads the bitmask to see which members the cert claims signed, looks up those members' real registered BLS keys via the committee (`Replicas::identities_by_rank`), aggregates them, and requires the result to equal the transmitted `public_key`. The committee is the only source of members' true keys, so without it the bind cannot run and nothing anchors the sender-supplied key to a real identity. (The separate weight check also consults the committee, but only to sum the weights of the bitmask-named members — which is why a forger names real members to clear the quorum threshold. Weight proves "enough members are named"; only the pubkey bind proves "those members signed.")

## The bug

That bind short-circuited to `Ok(())` whenever the verifier had no committee. The two timeout aggregators (global and app-shard) built their validator via the committee-less `new_with_timeout_domain`, and peer `TimeoutState`s route into them from the live inbound path (`message_loop`/`app_engine` -> `TimeoutProcessor::validate_timeout` -> `ConsensusValidator` -> this verifier). So an embedded QC/TC forgery — a self-signed `(pk, sig)` whose bitmask names real members to pass the weight check — was accepted on a reachable network path. The Go reference is not affected: its verifier always holds the active prover set and binds unconditionally.

## The fix

Make the committee non-optional so the fail-open is unrepresentable rather than guarded at runtime.

- `bls_verifier.rs`: `committee` becomes `Arc<dyn Replicas>` (no `Option`); the `None => Ok(())` arm is removed so the bind always runs; `verify_vote` no longer needs its committee-absent error path. The two committee-less constructors (`new`, `new_with_timeout_domain`) are removed and `new_with_committee` is renamed to `new` — now the only constructor, so every call site must supply a committee.
- `timeout_aggregation.rs` / `app_timeout_aggregation.rs`: build the verifier with the aggregator's committee (global uses an empty vote filter, app-shard the shard filter), matching the proposal gates.
- `app_engine.rs`, `archive_sync.rs`, `verify_migration.rs`, `validator.rs`: mechanical rename to the new `new`.

The newest-QC tracking order in `TimeoutProcessor` is deliberately left unchanged: track-before-verify is required for the TC-consistency invariant and matches Go, and making the timeout validators committee-aware already rejects a forged embedded QC before it is tracked.

## Testing

Besides the standard test suite, I also tested with 

```
cargo run -p devnet -- single --verbose --stopframe=5 \
  --rank-partitions='[{"rank":2 ,"partition1":["archive-1","archive-2"],"partition2":["archive-3", "archive-4"]}]'
```

(see #578 for devnet)

### DIFF
--- a/crates/quil-engine/tests/e2e_consensus.rs
+++ b/crates/quil-engine/tests/e2e_consensus.rs
@@ -3677,7 +3677,11 @@ async fn tier2_storage_audit_evicts_cheating_member() {
     );
     assert_eq!(before.kick_frame_number, 0, "precondition: not yet kicked");
 
-    let coverage_frame = build_global_frame_with_bundle(10, &bundle);
+    // STRICT LOCKSTEP: a storage frame's attestation must anchor to the
+    // IMMEDIATELY PRECEDING global frame, so the enclosing global frame number
+    // must be `global_frame_number + 1` (= 1001) to satisfy the
+    // `anchor == frame_number - 1` gate in `audit_storage_attestation`.
+    let coverage_frame = build_global_frame_with_bundle(1001, &bundle);
     let result = archive
         .materializer
         .materialize(&coverage_frame)
--- a/crates/quil-engine/src/app_engine.rs
+++ b/crates/quil-engine/src/app_engine.rs
@@ -1756,7 +1756,7 @@ impl AppConsensusEngine {
             // filter and the QC/TC `.filter()` match what voters signed.
             let raw_agg: Arc<dyn SignatureAggregator> =
                 Arc::new(BlsSignatureAggregator::new(bls_ctor.clone()));
-            let verifier = BlsConsensusVerifier::new_with_committee(
+            let verifier = BlsConsensusVerifier::new(
                 raw_agg,
                 vote_domain_for_validator,
                 timeout_domain_for_validator,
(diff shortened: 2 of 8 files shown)


crates/quil-engine/tests/e2e_consensus.rs
```diff
@@ -3677,7 +3677,11 @@ async fn tier2_storage_audit_evicts_cheating_member() {
     );
     assert_eq!(before.kick_frame_number, 0, "precondition: not yet kicked");
 
-    let coverage_frame = build_global_frame_with_bundle(10, &bundle);
+    // STRICT LOCKSTEP: a storage frame's attestation must anchor to the
+    // IMMEDIATELY PRECEDING global frame, so the enclosing global frame number
+    // must be `global_frame_number + 1` (= 1001) to satisfy the
+    // `anchor == frame_number - 1` gate in `audit_storage_attestation`.
+    let coverage_frame = build_global_frame_with_bundle(1001, &bundle);
```
dazthecorgi, line 3684: This just ports the fix from https://github.com/QuilibriumNetwork/monorepo/pull/577 to make tests pass

dazthecorgi:
superseded by 70465db4

PR #592, opened 2026-08-14
fix: serve app shard frames from worker-owned stores

### LINKED ISSUE #588: Improve app shard service reliability
One logical node is a master node and and some worker nodes. Worker nodes can be in process (thread workers) or a standalone worker process that the master communicates with via gRPC. Workers always have separate databases from the master.

The master node runs an app shard service that provides access to app shard frames via gRPC. App shard frames are normally stored in the worker databases. 

The app shard frames are currently duplicated in the master's database so that the master node can serve them via gRPC. If there is a failure the duplication is not retried, which means there can be gaps in the coverage of the app shard service. And if a node no longer serves a shard, the app frames in the master's store become orphaned.

I think it'd be better to avoid the duplication of data for reliability and consistency. In thread worker mode, the master could directly read the stores of the workers to serve app shard frame requests. In standalone worker mode, requests could be proxied.

blacks1ne:
## Summary

- serve AppShardService reads directly from the active local worker store or proxy them to the acknowledged active remote worker
- remove the master-side FullFrameProduced frame mirror that duplicated worker data and could leave gaps or orphaned frames
- retain authoritative archive-store fallback while preserving worker read and transport errors
- generation-guard local activation, serialize remote reassignment, send idle deallocation commands, and stop superseded app consensus hosts
- raise AppShardService transport limits to 64 MiB and preserve latest versus numbered frame semantics

## Validation

- Docker nextest focused suite: 45 passed, 629 skipped
- Docker cargo check --locked -p quil-node --tests
- Taskfile-equivalent AMD64 production node build
- regression coverage proves public reads succeed with the master store empty, plus archive fallback, reassignment ordering, deallocation, shutdown, error mapping, and responses above 4 MiB

Fixes #588

### DIFF
--- a/crates/quil-engine/src/cw_app_seams.rs
+++ b/crates/quil-engine/src/cw_app_seams.rs
@@ -106,7 +106,6 @@ pub fn app_frame_from_state(
         storage_attestation,
     }
 }
-
 /// Frame identity (`Poseidon(output)[..32]`) as the consensus digest — identical
 /// scheme to global (`AppShardState`/`GlobalFrame` share `compute_output_identity`).
 fn app_frame_digest(frame: &AppShardFrame) -> Option<Digest> {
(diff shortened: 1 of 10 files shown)


crates/quil-engine/src/cw_app_seams.rs
```diff
@@ -477,6 +476,16 @@ pub struct AppConsensusCwHandle {
...
+impl Drop for AppConsensusCwHandle {
+    fn drop(&mut self) {
+        // The commonware host runs independently of the app engine's tokio
+        // task. Ensure aborting/dropping that engine cannot leave an old-shard
+        // consensus host alive after reassignment or deallocation.
+        self.shutdown
+            .store(true, std::sync::atomic::Ordering::Release);
+    }
+}
+
```
dazthecorgi: Just for my understanding: is this related to https://github.com/QuilibriumNetwork/monorepo/issues/588?

PR #593, opened 2026-08-14
Fix duplicate app shard frame numbers

### LINKED ISSUE #591: Duplicate finalized frame numbers in app shard consensus
## Summary

App shard consensus can produce duplicate frame numbers which are silently dropped.

## Details

Simplex can enter view V+1 after V is notarized and certified but before V’s finalization has been applied to the shard clock store. Because the app proposer ignores the consensus parent and reads that store, V+1 can reuse V’s app frame number.

The worker persists the duplicate, skips materializing an already-seen number, then still republishes it and emits coverage at `crates/quil-engine/src/app_engine.rs:2429`. Archives silently ignore it at `crates/quil-engine/src/archive_ingest.rs:135`. A same-height clock-store entry may also be overwritten.

## Repro

I added a repro in this commit: https://github.com/dazthecorgi/monorepo/commit/7a15239535fc0bd87b9f394fcabb6556b3d6d897

blacks1ne:
## Summary

- build app-shard proposals from the exact parent selected by Simplex instead of rereading a potentially lagging clock-store head
- verify shard, view, height, and parent linkage before voting
- serialize lossless finalization delivery and suppress stale, conflicting, gapped, or replayed downstream publication and coverage
- key proposal request and attestation data by frame number plus Simplex rank

## Validation

- Docker nextest regression matrix: 6 passed, including one-prover and three-prover consensus and the issue reproduction
- broader consensus and store suite: 105 passed
- Linux/amd64 production quil-node release build succeeded

Closes #591

### DIFF
--- a/crates/quil-engine/src/app_engine.rs
+++ b/crates/quil-engine/src/app_engine.rs
@@ -339,7 +339,7 @@ struct AppLeaderProvider {
     /// `requests_root` is computed over these bundles' canonical
     /// encodings, so it is recomputable/verifiable from the frame.
     frame_requests: Arc<std::sync::Mutex<
-        std::collections::HashMap<u64, Vec<quil_types::proto::global::MessageBundle>>,
+        std::collections::HashMap<(u64, u64), Vec<quil_types::proto::global::MessageBundle>>,
     >>,
     /// KV backing the member's persisted PoRep replicas. Present iff this node
     /// participates in storage (built into a `ReplicaStore` in `prove_next_state`
@@ -353,7 +353,8 @@ struct AppLeaderProvider {
     /// `frame_requests`. Under commonware-simplex, votes carry no payload, so this
     /// is a PROPOSER SELF-attestation (single member = the frame's prover), NOT
     /// the legacy multi-member committee attestation assembled from vote openings.
-    frame_attestations: Arc<std::sync::Mutex<std::collections::HashMap<u64, Vec<u8>>>>,
+    frame_attestations:
+        Arc<std::sync::Mutex<std::collections::HashMap<(u64, u64), Vec<u8>>>>,
     /// Order-independent fingerprint of THIS CW instance's committee (the fixed
     /// simplex validator set built at `start_consensus_cw`). The finalization
     /// cert a proposed frame receives is signed by exactly this committee, but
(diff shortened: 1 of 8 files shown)


crates/quil-engine/src/app_engine.rs
```diff
@@ -584,32 +581,60 @@ impl quil_consensus::leader_provider::LeaderProvider<AppShardState> for AppLeade
...
+        //
+        // App-shard state roots are the materialized pre-state, so unlike the
+        // global chain we deliberately wait until the selected parent is both
+        // canonical and materialized. A NoVote only nullifies this view; once
+        // the finalization callback catches up, the next view can safely build
+        // N+1 from the exact parent bytes and identity.
+        let materialized = self
+            .shard_mat_frame
+            .load(std::sync::atomic::Ordering::SeqCst);
+        if materialized != prior_frame_number {
```
dazthecorgi, line 598: I'm not sure this is the right approach, because it creates an extra view per frame for app shard consensus.

Also, the `prior_frame_number` is attacker controlled, so trusting it here can cause liveness issues.

crates/quil-engine/src/cw_app_seams.rs
```diff
@@ -187,16 +211,27 @@ impl AppSeamProposer {
...
+                    view,
+                    parent = %hex::encode(&prior_state_id),
+                    "cw app propose: consensus parent metadata unavailable — nullify",
+                );
+                return None;
+            }
+        };
 
-        // App shards resolve their parent from the shard clock store internally,
-        // so `prior_frame_number` is advisory; `prior_state_id` is the identity.
```
dazthecorgi, line 199: I think the "`prior_frame_number` is advisory" part is important to preserve and maybe make more explicit. If I understand correctly, it refers to the fact that the prior frame number is learned via gossip and never verified. An attacker can send an invalid frame number over gossip causing a halt.

crates/quil-engine/src/cw_app_seams.rs
```diff
@@ -228,6 +269,38 @@ impl GlobalProposer for AppSeamProposer {
...
+                    "cw app verify: consensus parent metadata unavailable — nullify",
+                );
+                return false;
+            }
+        };
+        if !app_header_extends_simplex_parent(
+            header,
+            &self.filter,
+            view,
+            parent_frame_number,
```
dazthecorgi, line 293: The parent frame number is attacker controlled and trusting it for verify can cause liveness issues.

crates/quil-engine/src/cw_app_seams.rs
```diff
@@ -187,16 +211,27 @@ impl AppSeamProposer {
...
+        {
+            Some(frame_number) => frame_number,
+            None => {
+                tracing::warn!(
+                    view,
+                    parent = %hex::encode(&prior_state_id),
+                    "cw app propose: consensus parent metadata unavailable — nullify",
+                );
+                return None;
+            }
```
dazthecorgi, line 229: I think we should fall back to the store here (similar to the global propose implementation).

PR #633, opened 2026-08-28
feat: show app shard materialization health in TUI

blacks1ne:
Adds compact Mat, Lag, and State fields to the prover-manage TUI. The archive protocol reports the durable per-shard materialization cursor and latest retained app-frame height; Lag is derived from those same-shard values, never the global head. Existing archives remain wire-compatible and display Unknown until they expose the new fields.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -101,6 +101,15 @@ debug = false
 # and overflow checks remain on. `cargo build`/`cargo check` are
 # unaffected — they use profile.dev.
 
+# Test binaries keep file/line for every frame — backtraces, panic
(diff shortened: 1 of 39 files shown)


blacks1ne:
Validation note: the Available-panel unit/cursor regression was introduced only in the diagnostic-build reconciliation, not in this PR branch. #633 already inherits #585’s cursor-independent bare Size [MB] and Reward [Q/f] cell rendering, so no #633 force-push is needed. The diagnostic correction adds a focused regression test; it passed on l1 (1 passed, 0 failed).

